PR #8, opened 2017-06-02
kit tests in make test

kegl:


agramfort:
LGTM

.travis.yml
```diff
@@ -6,6 +6,7 @@ before_install:
     - export PATH=/home/travis/miniconda/bin:$PATH
     - conda update --yes --quiet conda
 install:
+    - pip install codecov
```
agramfort: do the pip install after the conda install

.travis.yml
```diff
@@ -6,6 +6,7 @@ before_install:
...
     - conda create -n testenv --yes pip python=3.6
     - source activate testenv
     - conda install --yes --quiet numpy scipy matplotlib scikit-learn pandas nose
```
agramfort: pip install codecov

here

.travis.yml
```diff
@@ -17,12 +18,14 @@ install:
     - git clone https://github.com/ramp-kits/iris.git
     - cd ..
 script:
-    - nosetests rampwf/tests
+    - nosetests --with-coverage rampwf/tests
     - cd kits/boston_housing
```
agramfort: don't move around with cd as coverage will be put the current dir so it will not be in a single place

kegl: done

PR #11, opened 2017-06-03
[WIP] testing backend submission

glemaitre:


agramfort:
@kegl review time :)

kegl:
Let's deal with this after the zip encrypting is solved.

kegl:
I think I merged too fast. I don't see /bins/test_submission

jorisvandenbossche:
Usually we use the convention that "WIP" in the title means it is work in progress and should not yet be merged (but you eg already want to test it or that other people can look at it)

PR #14, opened 2017-06-20
Titanic

kegl:
Please gimme a fast review so I can make titanic green again.


agramfort:
merge if you're happy

PR #22, opened 2017-06-21
[MRG] Refactoring testing

glemaitre:
Do not merge it yet ... this is experimental ;)

PR #24, opened 2017-06-21
[MRG] add command line module

### LINKED ISSUE #16: Make executable name more explicit
The current executable name is called`test_submission` when `ramp-workflow` is installed on the system.

I think the name should at least be related to `ramp` or `ramp-workflow` to avoid confusion and to keep the namespace consistant.

glemaitre:
Closes #16 

Add a command line module and modify the `setup.py`. It can be called with `ramp_test_submission --ramp_kit_dir --ramp_data_dir --submodule_name` with the same default as previously done.

#22 needs to be merged first and a merge to master need to be applied


rampwf/utils/command_line.py
```diff
@@ -0,0 +1,35 @@
...
+    parser = argparse.ArgumentParser(
+        prog='ramp_test_submission',
+        description='Test your ramp-kit before attempting a submission.')
+    parser.add_argument('--ramp_kit_dir',
+                        default='.',
+                        nargs=1,
+                        type=str,
+                        help='Root directory of the ramp-kit to test.')
+    parser.add_argument('--ramp_data_dir',
+                        default='.',
```
jorisvandenbossche: shouldn't the default be `./data` ?

rampwf/utils/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .command_line import ramp_test_submission
```
jorisvandenbossche: I don't think it is needed to import this here? (for the console script, you access it from utils.command_line no?)

glemaitre:
It is green @jorisvandenbossche

PR #29, opened 2017-06-22
FIX command line keyword arguments

jorisvandenbossche:


jorisvandenbossche:
@kegl this should fix it

I will still add test

PR #31, opened 2017-06-22
CMD rename submission_name to shorter submission

jorisvandenbossche:


jorisvandenbossche:
merged since I need this to update the kits

PR #34, opened 2017-06-23
[MRG] Replace bundle list with table

aboucaud:
@kegl use this PR instead


glemaitre:
Before merging could you make the markdown use 79 characters max.
It is easier to review and to render in some old fashion editors :D

aboucaud:
How do you do that ? It affects the rendering if you add newlines.

PR #35, opened 2017-06-23
Fix 'make test' + switch to pytest

jorisvandenbossche:


jorisvandenbossche:
Merging, as @kegl wants make test fixed :-)

PR #36, opened 2017-06-25
Epidemium

kegl:


kegl:
Can we merge this?

jorisvandenbossche:
What is the `n_colunms` about?

kegl:
n_columns is the number of columns of the output y_pred (or y_true). I do this assert because submitters (RAMP participants) can return whatever they want in their last predict(), and I want to throw an error early error if the format of their y_pred is illegal.

jorisvandenbossche:
Ah, OK, that makes sense. We only should then give a more informative error message to the user.

kegl:
True. You want to give it a try?

jorisvandenbossche:
Yes, but not today :-)

PR #38, opened 2017-07-02
The new way of creating Predictions.

kegl:
The goals with this PR:
1. Making predictions types (`Predictions`) more transparent.
2. Eliminating the need to put `label_names` each time when we create a new object.
3. Preparing combined predictions (e.g. in drug spectra) when `y_pred` is a concatenation of different `Predictions` (e.g. a regression and a classification).
4. Making it easier and more transparent to score combined predictions (e.g. in drug spectra).

I needed to make `label_names` a class variable, and find a way to return a class with _fixed_ `label_names` in the setup scripts `problem.py`. This required creating classes on the fly with `type()` in `prediction_types.regression` and `prediction_types.multiclass`.

Drug spectra (combined predictions and scores) is not yet in the PR but it will be way more simple and transparent than now. Moreover, it will be way easier to create other workflow types with mixed predictions.

Also note that some of the features (those in `prediction_types.base`) are not yet tested in rampwf only in ramp-board (cv bagging and model combination), so those functions may seem useless here. I will bring it to`testing.py` once the dust settles (I port the production site to the new setup).


rampwf/prediction_types/multiclass.py
```diff
@@ -12,61 +12,57 @@
...
+@property
+def _y_pred_label_index(self):
+    """Multi-class y_pred is the index of the predicted label."""
+    return np.argmax(self.y_pred, axis=1)
 
-    @property
-    def y_pred_label(self):
-        return self.labels[self.y_pred_label_index]
+
+def make_predictions_type(label_names=[]):
```
agramfort: label_names suggests it's only for classification. If so it should be explicit in the function name

rampwf/prediction_types/regression.py
```diff
@@ -10,17 +10,28 @@
...
-                             "or n_samples")
+            shape = (n_samples, n_columns)
+        self.y_pred = np.empty(shape, dtype=float)
+        self.y_pred.fill(np.nan)
+    else:
+        raise ValueError(
+            'Missing init argument: y_pred, y_true, or n_samples')
+
+
+def make_predictions_type(label_names=[]):
```
agramfort: this one is regression?

why do you need so many functions with the same name?

rampwf/prediction_types/multiclass.py
kegl: I do use it for regression, e.g. in epidemium, where there are 6 targets. My semantics is that label_names name the columns of y_pred, when y_pred is 2D. It is [] when it is 1D. So len(label_names) is the second dim of y_pred.

rampwf/prediction_types/regression.py
kegl: We can change this later. It's just to have a single syntax when we use it, e.g. 
https://github.com/ramp-kits/titanic/blob/master/problem.py

Predictions = rw.prediction_types.multiclass.make_predictions_type(
    label_names=_prediction_label_names)

It could be simplified, e.g. if we expose them in __init__ to

Predictions = rw.prediction_types.make_regression_type(
    label_names=_prediction_label_names)

Would you prefer this?

kegl: or make_regression

kegl:
I finally figured out the transparent way of handling RAMPs with multi-output y_pred. It my seem a bit convoluted (check score_types/combined and score_types/make_combined), but most importantly look at the problem description script 
```
https://github.com/ramp-kits/drug_spectra/blob/master/problem.py
```
and check whether the way to use this feature is clear. 

cc @aboucaud in case we end up having multi-target RAMPs in astro.

PR #41, opened 2017-07-13
[MRG] El nino

kegl:
This is the most complicated WF because we relax the usual condition that X and y have the same length.


agramfort:
you have a bunch of lines not covered by any test. See codecov report.

kegl:
It's mostly trivial.
1) clustering is not done yet
2) some features are only tested in ramp-board for now
3) I don't know what is the ergonomic way to test errors, that is, when the normal functionality is tho throw an error. E.g., https://github.com/ramp-kits/el_nino/tree/master/submissions/illegal_lookahead could be called easily but it crashes travis since it's normal behavior is to crash. If some of you shows me an example of how to deal with this, I can take care all the raises.

kegl:
The git lfs/netcfd/travis bug is a bigger concern now, since that combination can become quite common for large time series RAMP, and currently we can't use travis to test them. See the slack.

agramfort:
use assert_raises to test that an exception is raised

and mock the data to make a test.

PR #42, opened 2017-07-19
Tracking

kegl:
* adding clustering stuff (for HEP_tracking, tested by HEP_tracking)
* fixing testing: ramp_test_submission is used by users, no need to test the notebook, so notebook testing (used by ramp builders) is a separate script now


kegl:
I'm merging this so I can proceed with testing the kits, but keep commenting.

PR #46, opened 2017-08-07
Image classification

kegl:


jorisvandenbossche:
@kegl did someting go wrong here? As the diff of this PR is just one line of whitespace change

kegl:
Not sure. @mehdidc  committed on the image_classification branch after I merge it, so I reopened the PR and remerged, but it's possible I made a mess. Master seems ok now.

PR #47, opened 2017-09-12
Score and feature extraction additions for California Rainfall hackathon

djgagne:
I have added Brier Score scoring options and a gridded feature extractor for processing netCDF files with xarray that contain multiple ensemble members.

### DIFF
--- a/rampwf/score_types/brier_score.py
+++ b/rampwf/score_types/brier_score.py
@@ -1,0 +1,31 @@
+from .base import BaseScoreType
+import numpy as np
+
+
+class BrierScore(BaseScoreType):
+    is_lower_the_better = True
+    minimum = 0.0
+    maximum = 1.0
+
+    def __init__(self, name='brier_score', precision=2):
+        self.name = name
+        self.precision = precision
+
+    def score_function(self, ground_truths, predictions, valid_indexes=None):
+        """A hybrid score.
+
+        It tests the the predicted _probability_ of the second class
+        against the true _label index_ (which is 0 if the first label is the
+        ground truth, and 1 if it is not, in other words, it is the
+        tru probabilty of the second class). Thus we have to override the
+        `Base` function here
+        """
+        if valid_indexes is None:
+            valid_indexes = slice(None, None, None)
+        y_proba = predictions.y_pred[valid_indexes][:, 1]
+        y_true_proba = ground_truths.y_pred_label_index[valid_indexes]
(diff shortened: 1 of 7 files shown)


kegl:
Please add the new starting kit to travis to test the workflow and the scores.

agramfort:
why not using the brier_scrore from sklearn?

agramfort:
see http://scikit-learn.org/stable/modules/generated/sklearn.metrics.brier_score_loss.html

djgagne:
I used my own implementation of Brier Score instead of scikit-learn's since it is very simple to implement (essentially just mean squared error for probabilities). I can replace my implementation with the one in scikit-learn if the team prefers that, but there should not be any differences in performance or results.

kegl:
Yes please. Accuracy or error are even simpler, yet we like to rely on sklearn implementation.

djgagne:
Because Travis CI is timing out on downloading the files for the california_rainfall hackathon, I created a separate [california_rainfall_test](https://github.com/djgagne/california_rainfall_test) repo that uses much smaller versions of the Hackathon files. I would like to have it included in the ramp-kits organization so that it can be tested by ramp-workflow.

kegl:
Please just add a top level comment to the workflow so people know what kind of setup it can be used for (beyond this challenge). Something we did in https://github.com/paris-saclay-cds/ramp-workflow/blob/master/rampwf/workflows/el_nino.py. Otherwise LGTM, @agramfort?

djgagne:
I added a time dimension name lookup in grid_feature_extractor.py so that the code would be a little more generalizable in addition to the top level comments. There was a minor bug in my implemented that I addressed.

rampwf/score_types/brier_score.py
```diff
@@ -0,0 +1,154 @@
+from .base import BaseScoreType
+import numpy as np
+from sklearn.metrics import brier_score_loss
+
+class BrierScore(BaseScoreType):
+    is_lower_the_better = True
+    minimum = 0.0
+    maximum = 1.0
+
+    def __init__(self, name='brier_score', precision=3):
```
agramfort: please add proper docstrings for the functions, methods and classes.

Here it should be

"""
Parameters
------------
name : str
     The human readable name for the score.
precision : int
     The floating point precision of ???
"""

we follow numpydoc formatting guidelines.

see Extended Summary section in : https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

PR #48, opened 2017-09-15
Add current directory to python path when running test_submission

jorisvandenbossche:
This allows that local packages in the starting kit are importable

That is what we are doing for now in the https://github.com/ramp-kits/mars_craters starting kit, where there is a `workflow` directory with all the custom code we need for defining the workflow in problem.py (it is too much too put eg everything in problem.py)

### DIFF
--- a/rampwf/utils/testing.py
+++ b/rampwf/utils/testing.py
@@ -12,6 +12,10 @@ import numpy as np
 import cloudpickle as pickle
 
 
+import sys
+sys.path.append('')
+
+
 def _delete_line_from_file(f_name, line_to_delete):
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
I have to say that this is a bit of a hack, but the whole test_submission that imports `problem.py` using the `imp` module is somewhat hacky anyway.

kegl:
Fine with me. I know it's hacky, feel free to try to find a better solution. I actually use `imp` partly so that we can import files directly, without `.` being a module. So I think using `imp` also to import the workflow would also work.

kegl:
#279 solves this in an alternative way.

PR #49, opened 2017-09-15
add transform_test to ImageClassifier

mehdidc:


jorisvandenbossche:
What is the use case of having a different transform method for train vs test?

And if we are adding this, can you add it somewhere to the ImageClassification docstring to specify that the `image_preprocessor` file also needs to implement a `transform_test` method ?

kegl:
+1 for adding it to the docstring

kegl:
The use case is that you may want to e.g. have a random rotation when training (for increasing effective data size), but not at test time (it would randomize the test score). This actually happened at last challenge. It's not a big deal, but we thought we would let them decouple the two.

kegl:
@mehdi: instead of upgrading all the starting kits, I think transform_test should be optional, in the workflow. You should catch when it's not there and fall back to transform. If we don't do it, we break backward compatibility (all previous submissions on the site will fail)

mehdidc:
Yes, it's true, I was thinking that it's perhaps better to make it explicit (transform_test), but adding the docstring should be enough to document this feature

kegl:
LGTM.

rampwf/workflows/image_classifier.py
```diff
@@ -42,18 +79,22 @@ def train_submission(self, module_path, folder_X_array, y_array,
         image_preprocessor = imp.load_source(
             '', submitted_image_preprocessor_file)
         transform_img = image_preprocessor.transform
+        has_transform_test = hasattr(image_preprocessor, 'transform_test')
+        transform_test_img = image_preprocessor.transform_test if has_transform_test else transform_img
```
jorisvandenbossche: I think you can also do `transform_test_img = getattr(image_preprocessor, 'transform_test', transform_img)`

jorisvandenbossche:
Small comment, but looks good! Indeed better to fall back

PR #50, opened 2017-09-15
Making ramp_workflow pip installable

aboucaud:
This PR intends to clean up the dependency tree of ramp_workflow.

The idea is to trim off all dependencies related to a particular ramp-kit.

The ramp-kits will therefore need to have a specific requirement.txt file listing the dependencies, which will be installed then using pip or conda.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: python
 dist: trusty
 env:
-  - PYTHON_VERSION=2.7
-  - PYTHON_VERSION=3.5
-  - PYTHON_VERSION=3.6
+  - PYTHON_VERSION=2.7 IPYTHON_KERNEL=python2
+  - PYTHON_VERSION=3.5 IPYTHON_KERNEL=python3
+  - PYTHON_VERSION=3.6 IPYTHON_KERNEL=python3
 before_install:
     - wget -q http://repo.continuum.io/miniconda/Miniconda-3.6.0-Linux-x86_64.sh -O miniconda.sh
     - chmod +x miniconda.sh
@@ -14,6 +14,7 @@ install:
     - conda create -n testenv --yes pip python=$PYTHON_VERSION
     - source activate testenv
     - pip install -r testing-requirements.txt
+    # Adding tensorflow here due to failures in Py2.7 build fails with MNIST
     - pip install tensorflow
     - pip install .
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,8 @@ before_install:
 install:
     - conda create -n testenv --yes pip python=$PYTHON_VERSION
     - source activate testenv
-    - conda install --yes pandas scikit-learn scikit-image cloudpickle joblib gitpython
-    - conda install --yes pytest flake8 nbconvert jupyter
+    - pip install -r testing-requirements.txt
+    - pip install tensorflow
     - pip install .
 script:
     - pytest -s -v rampwf
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,8 @@ install:
     - conda create -n testenv --yes pip python=$PYTHON_VERSION
     - source activate testenv
     - source install_requirements.sh
-    - pip install pytest pytest-cov codecov
-    - pip install -q flake8
+    - pip install -q pytest pytest-cov codecov
+    - pip install -q flake8 gitpython jupyter
     - pip install .
 script:
     - pytest -s -v --cov=rampwf rampwf
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,5 +19,7 @@ install:
     - pip install .
 script:
     - pytest -s -v --cov=rampwf rampwf
+after_success:
+    - codecov
 notifications:
   email: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ install:
     - conda create -n testenv --yes pip python=$PYTHON_VERSION
     - source activate testenv
     - source install_requirements.sh
-    - pip install codecov
+    - pip install pytest pytest-cov codecov
     - pip install -q flake8
     - pip install .
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,6 @@ install:
     - pip install tensorflow
     - pip install .
 script:
-    - pytest -s -v rampwf
+    - pytest -s -v --cov=rampwf rampwf
 notifications:
   email: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,11 +13,10 @@ before_install:
 install:
     - conda create -n testenv --yes pip python=$PYTHON_VERSION
     - source activate testenv
-    - pip install -r testing-requirements.txt
+    - conda install --yes pandas scikit-learn scikit-image cloudpickle joblib gitpython
+    - conda install --yes pytest flake8 nbconvert jupyter
     - pip install .
 script:
-    - pytest -s -v --cov=rampwf rampwf
-after_success:
-    - codecov
+    - pytest -s -v rampwf
 notifications:
   email: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,7 @@ before_install:
 install:
     - conda create -n testenv --yes pip python=$PYTHON_VERSION
     - source activate testenv
-    - source install_requirements.sh
-    - pip install -q pytest pytest-cov codecov
-    - pip install -q flake8 gitpython jupyter
+    - pip install -r testing-requirements.txt
     - pip install .
 script:
     - pytest -s -v --cov=rampwf rampwf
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -13,9 +13,7 @@ before_install:
 install:
     - conda create -n testenv --yes pip python=$PYTHON_VERSION
     - source activate testenv
-    - source install_requirements.sh
-    - pip install codecov
-    - pip install -q flake8
+    - pip install -r testing-requirements.txt
```
jorisvandenbossche: since we are using conda envs, let's install with conda here?

setup.py
```diff
@@ -34,35 +34,41 @@
...
+            'Programming Language :: Python',
+            'Topic :: Software Development',
+            'Topic :: Scientific/Engineering',
+            'Operating System :: Microsoft :: Windows',
+            'Operating System :: POSIX',
+            'Operating System :: Unix',
+            'Operating System :: MacOS'],
+        install_requires=[
+            'numpy>=1.13',
+            'scipy>=0.19',
```
jorisvandenbossche: scipy is not needed I think?

setup.py
```diff
@@ -34,35 +34,41 @@
...
+            'scikit-learn>=0.18'],
```
jorisvandenbossche: although pandas is not a strict requirement (it is currently not directly imported), it is still used a lot in the code (code that assumes pandas data structures are used), so therefore would add it here

jorisvandenbossche:
I also think it might be better to split the requirements parsing into another PR, so the actual setup.py updates can already be merged (or, for now still install some of the extra deps in travis.yml until all starting_kits have gotten a requirements file)

setup.py
aboucaud: indeed, but scikit-learn expects it and it is not installed as dependency

.travis.yml
aboucaud: What is the command ? I don't use conda..

aboucaud:
I pushed a `requirements.txt` in every starting kit tested so far

setup.py
aboucaud: fair enough

aboucaud: what should be the version required ?

.travis.yml
aboucaud: ok, bad idea. If a single library is not installable by conda the whole thing crashes.

rth: Maybe use an `environement.yml` file then (example [here](https://github.com/binder-project/example-conda-environment/blob/master/environment.yml)) that can be used to update an existing conda environment and can specify both conda and pip packages ? 

Also as a side note, I find https://github.com/astropy/ci-helpers very useful for setting up continuous integration.

setup.py
rth: Maybe better to read from `requrements.txt` here? If you install scikit-learn with conda, scipy will get installed... Also, joblib is imported but is not in the requirements...

setup.py
jorisvandenbossche: Hmm, not sure. Let's maybe take 0.19 for now ? 
(actually, we should in principle also test for all minimal dependencies, but can be for other PR)

.travis.yml
jorisvandenbossche: Which one is not installable? All in that list are for me installable with conda

jorisvandenbossche: Ah, but that is probably because I have conda-forge channel added. OK, then leave it be.

Regarding jupyter, what is this used for? Is it only for nbconvert? (then can also only install that)

aboucaud:
Only the python2 build fails because of the following import error

```
from tensorflow.core.framework.graph_pb2 import *
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
    import sys
    _b=sys.version_info[0]<3 and (lambda x:x) or (lambda x:x.encode('latin1'))
>   from google.protobuf import descriptor as _descriptor
E   ImportError: No module named google.protobuf

/home/travis/miniconda/envs/testenv/lib/python2.7/site-packages/tensorflow/core/framework/graph_pb2.py:6: ImportError
```
This is specific to tensorflow and could maybe be solved easily by installing the `protobuf` lib, but since the former builds seems to pass and you did not have that dependency, I find it weird.

Any idea @kegl ?

setup.py
aboucaud: I agree its a duplication. But I don't know what is best regarding pypi.

Concerning joblib, I moved the imports to get rid of the dependency.

.travis.yml
aboucaud: @rth  How do you use an `environment.yml` to specify pip packages ?

IMO `ci-helpers` is quite a big machinery for pure python packages. However we may use it in the future, especially for the very simple way to define build matrices.

kegl: Yes, jupyter is only for nbconvert, I only call it in the backend to generate the page on ramp.studio. So it should be tested in the travis of the kit, but not in rampwf.

kegl:
Have never seen it. Pinging @mehdidc :)

.travis.yml
rth: @aboucaud Well one could do, `conda env create [or update] -f envonement.yml` where, `environement.yml` looks like, e.g.,
```yml
name: test-env
channels:
  - defaults
  - conda-forge
dependencies:
  - python=3.5
  - numpy=1.13.1
  - scipy=0.19.1
  - conda_package_1
  - conda_package_2
  - pip :
    - pip_package_1 
    - ./  # install the current package
```
but in the end it's not that different from just running conda install and pip install separately..

setup.py
jorisvandenbossche: let's leave them separate, the small duplication is not that bad. There is some slight ideological difference between requirements.txt (deploy specific versions) and install_requires (specify minimum dependencies the deployed environment should adhere to). See https://packaging.python.org/discussions/install-requires-vs-requirements/

jorisvandenbossche:
> This is specific to tensorflow and could maybe be solved easily by installing the protobuf lib, but since the former builds seems to pass and you did not have that dependency, I find it weird.

Strange thing is that `protobuf` is actually installed in the failing travis build: https://travis-ci.org/paris-saclay-cds/ramp-workflow/jobs/276098845#L1314.

.travis.yml
aboucaud: Seems convenient indeed. I'll keep that in mind.

aboucaud:
Mmmmh I find several complains online regarding `protobuf` and `tensorflow` but only fixups, no actual recipe to make it work every time. 

Both libraries versions seem to be tightly linked so what I can think of is:
* protobuf released a new version 14h ago (see [here](https://github.com/google/protobuf/releases)) so it could just be bad luck that a build or a wheel was updated for py3.x but not py2.7 on pypi => the build may pass or fail erratically
* we should (as was said earlier) specify versions for these particular packages. Especially the ones with incompatible API's (like `keras` of `tf`). @mehdidc & @kegl, what versions do you usually run MNIST with ?

mehdidc:
@aboucaud in pollenatin insects 3, we use tensorflow 1.3.0 and protobuf 3.4.0

aboucaud:
The MNIST ramp-kit build and the ramp-workflow build are identical but one fails and the other passes..
* https://travis-ci.org/ramp-kits/MNIST/jobs/276042142#L915
* https://travis-ci.org/paris-saclay-cds/ramp-workflow/jobs/276098827#L1307

Now I'm really confused.. 😩

jorisvandenbossche:
Maybe try to list protobuf explicitly in the requirements for MNIST ? (don't know why that would matter, but in all the issues/questions about this issue online I saw a lot of 'solution' (like just reinstalling) for which I don't know why it would matter ...)

aboucaud:
At this point I really don't understand the difference between the two builds (MNIST alone or called by the ramp-workflow tests) on python2.7
Pinging @glemaitre @agramfort for new insights..

jorisvandenbossche:
Comaparing the two last builds, there is a difference in the protobuf version (3.3 vs 3.4)

jorisvandenbossche:
Also, to really see whether the same works on MNIST, I think we need to adapt the travis.yml file on MNIST to use the requirements.txt file

aboucaud:
> Comaparing the two last builds, there is a difference in the protobuf version (3.3 vs 3.4)

I tried this change this morning to see if that specific version was troubling. Apparently not.

jorisvandenbossche:
> Also, to really see whether the same works on MNIST, I think we need to adapt the travis.yml file on MNIST to use the requirements.txt file

Trying that in https://github.com/ramp-kits/MNIST/pull/3

jorisvandenbossche:
@aboucaud you can remove that last commit again. I just tried to make it even more comparable to how the env is set up in the MNIST repo, but also this attempt still fails with the same problem.

PR #53, opened 2017-09-20
Add flake8 automatic checking

aboucaud:


kegl:
I prefer this style

```
pos_obs_freq = np.histogram(
    y_proba[y_true_proba == 1], bins=self.bins)[0]
```

instead of this

```
pos_obs_freq = np.histogram(y_proba[y_true_proba == 1],
                            bins=self.bins)[0]
```

Issue #54, opened 2017-09-23
Python setup.py install failed


vfdev-5:
Hi,

I try to install github master version on a Ubuntu machine with command:
```
python3 setup.py install
```
Here is error I get:
```
Traceback (most recent call last):
  File "setup.py", line 47, in <module>
    long_description=open('README.md').read(),
  File "/usr/lib/python3.5/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 10306: ordinal not in range(128)
```
If I change 
```
long_description="",
```
installation succeeds. 

Cheers

PR #55, opened 2017-09-23
Update setup.py

vfdev-5:
Fix [#54](https://github.com/paris-saclay-cds/ramp-workflow/issues/54)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ if __name__ == "__main__":
         url=URL,
         version=VERSION,
         download_url=DOWNLOAD_URL,
-        long_description=open('README.md').read(),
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Thanks!

While you are at it, could you maybe change it slightly further to actually close the file again (to not leave open file handlers):

https://github.com/pypa/sampleproject/blob/a9c1287d5f9825d8d8fce32f7ab4bc7813b97179/setup.py#L17-L18

vfdev-5:
Nice example! Fixed as you asked. Let's wait for checks to be completed...

aboucaud:
Thanks for fixing this.

PR #64, opened 2017-10-14
Fixing Minipatch detection

aboucaud:


rampwf/score_types/detection.py
```diff
@@ -239,55 +214,99 @@ def scp_single(y_true, y_pred, shape, minipatch=None):
...
-def ospa_single(y_true, y_pred, cut_off=1, minipatch=None):
+def ospa(y_true, y_pred, minipatch=None):
+    """
+    Optimal Subpattern Assignment (OSPA) metric for IoU score.
+
+    This metric provides a coherent way to compute the miss-distance
+    between the detection and alignment of objects. Among all
+    combinations of true/predicted pairs, if finds the best alignment
+    to minimise the distance, and still takes into account missing
+    or in-excess predicted values through a cardinality score.
```
jorisvandenbossche: Can you add this docstring (also) to the OSPA score type class ?

rampwf/score_types/detection.py
```diff
@@ -239,55 +214,99 @@ def scp_single(y_true, y_pred, shape, minipatch=None):
...
     n_pred = len(y_pred)
-
-    n_max = max(n_true, n_pred)
-    n_min = min(n_true, n_pred)
+    n_total = n_true + n_pred
 
     # No craters and none found
     if n_true == 0 and n_pred == 0:
-        return 0
+        return 2, 2, 2
```
jorisvandenbossche: Is this correct? As that would give an OSPA of 2\*2 / 2 = 2, and thus score of 1 - 2 = -1 for a single patch? 
(didn't think this through yet, just noticed the consequence of those values)

rampwf/score_types/detection.py
```diff
@@ -239,55 +214,99 @@ def scp_single(y_true, y_pred, shape, minipatch=None):
...
-        penalizing value for wrong cardinality
     minipatch : list of int, optional
         Bounds of the internal scoring patch (default is None)
 
     Returns
     -------
-    float: distance between input arrays
+    (iou_sum, n_pred, n_total)
+        float - sum of ious of matched entries
+        int - sum of matched entries
```
jorisvandenbossche: sum -> number ? (and below as well)

rampwf/score_types/detection.py
jorisvandenbossche: but maybe a score on a single patch for this (no craters, non found) will maybe never make much sense.

The question is how to weight the fact this was done correctly (detecting there are no craters on this particular patch). When taking the aggregate of all patches, maybe we shouldn't weight this and just return 0,0,0 ?

jorisvandenbossche:
A general thought, can we stop calling this 'minipatch' ? I would find something like `score_area` much clearer

rampwf/score_types/detection.py
kegl: Yes, I would return 0, 0, 0

rampwf/score_types/detection.py
aboucaud: I don't get your comment

rampwf/score_types/detection.py
aboucaud: Sure thing

aboucaud: Done

aboucaud:
IMO the name should reflect the content. Since it is a bounding box, or the opposite corners of a rectangle, I would not call that an `area`.

rampwf/score_types/detection.py
jorisvandenbossche: Meaning of my cryptic comment: "I *think* 'sum' should be replaced by 'number' in that line"

Just because a "sum of entries" is a bit a strange thing to calculate, you cannot really sum them. But you can count them. Therefore I suggested "number".

jorisvandenbossche:
Forgot to push the comment button monday:

> Though from a marketing point of view minipatch is like 100x better. People would remember it. Like minibatch vs stepping_batch.

I don't know what stepping_batch is, but IMO minipatch and minibatch are not really comparable. 
Our patches are already 'mini' versions of the big image, therefore I find this a confusing terminology.

> IMO the name should reflect the content. Since it is a bounding box, or the opposite corners of a rectangle, I would not call that an area

For me a bounding box is a description of an area :-) (area is not only the measurement, but also a region) But `scoring_region` is also a possibility.

rampwf/score_types/detection.py
```diff
@@ -86,28 +87,48 @@ def detection_score(self, y_true, y_pred):
...
+    ---------
+    name : str, optional
+        Method name
+    precision : int, optional
+        Rounding precision for the score (default is 2)
+    conf_threshold : float, optional
+        Confidence threshold value use for the Average Precision
+        measurement (default is 0.5)
+    minipatch : list of int, optional
+        Bounds of the internal scoring patch (default is None)
```
jorisvandenbossche: Can you be more explitict here in the format ? (row_min, row_max, col_min, col_max ?)

rampwf/score_types/detection.py
```diff
@@ -310,15 +374,20 @@ def _select_minipatch_tuples(y_list, minipatch):
...
+    y_list_cut = ((y_list[0] >= row_min) & (y_list[0] < row_max) &
+                  (y_list[1] >= col_min) & (y_list[1] < col_max))
 
     return y_list_cut
 
 
-def _match_tuples(y_true, y_pred, minipatch=None):
+def filter_minipatch_tuples(y_list, minipatch):
+    in_minipatch = _select_minipatch_tuples(y_list, minipatch)
+    return np.array(y_list)[in_minipatch].tolist()
```
jorisvandenbossche: this could also be:

```
 [obj for (obj, cond) in zip(y_list, in_minipatch) if cond]
```

which preserves the list of tuples, and is for small lists (which is always the case since we are inside a single patch) also much faster (no numpy overhead)

rampwf/score_types/detection.py
```diff
@@ -438,14 +504,20 @@ def precision(y_true, y_pred, matches=None, iou_threshold=0.5,
...
         Threshold to determine match
+    minipatch : list of int, optional
+        Bounds of the internal scoring patch (default is None)
 
     Returns
     -------
     precision_score : float [0 - 1]
     """
+    if minipatch is not None:
+        y_pred = [filter_minipatch_tuples(y_patch, minipatch)
```
jorisvandenbossche: Can you add something like the following

```
# precision is calculated relative to the number of predicted objects, 
# so we filter for the scoring region based on the predictions  
```

to clarify the reasoning of the code

(the same for recall)

PR #66, opened 2017-10-16
Updated average precision calculation

jorisvandenbossche:
(still need to clean it up and add tests)

From a quick benchmark, this improves the performance considerably and makes it similar to the other scores in time to calculate:

- for 500 patches: 1.1 s to 100 ms (while precision or recall take ca 40 ms, SCP 200 ms)
- on the full test data of 10000 patches: 23s -> 2s (while precision or recall take ca 0.7 s)

If needed this could still be improved (as for the lower confidence values, it is not really needed anymore to do it step by step for a single new prediction, as this has only minor impact on the final average precision), but I don't think that is needed.

Regarding the precision. For the example with 500 test patches, I compared with the "fully exact" method (so for each step in including the next prediction, recalculate the full exact match), the two methods are *very* similar: 0.55176 (the full exact) vs 0.55224 (the 'greedy', this PR). So only a minor difference. 
Note that on current master, we use not a "full exact", but only an exact match for a certain number of confidence thresholds, and this gives 0.5254 (so less exact than the 'greedy' method in this PR anyhow, and much slower)


rampwf/score_types/detection.py
```diff
@@ -144,14 +144,9 @@ def __init__(self, name=_name, precision=2, iou_threshold=0.5):
...
-                       for single_detection in y_pred
-                       for detected_object in single_detection]
-        min_conf, max_conf = np.min(y_pred_conf), np.max(y_pred_conf)
-        conf_thresholds = np.linspace(min_conf, max_conf, 20)
-        ps, rs = precision_recall_curve(y_true, y_pred, conf_thresholds,
-                                        iou_threshold=self.iou_threshold)
-        return average_precision_interpolated(ps, rs)
+        ps, rs = precision_recall_curve_greedy(
+            y_true, y_pred, iou_threshold=self.iou_threshold)
+        return average_precision_exact(ps, rs)
```
aboucaud: `precision_recall_curve_greedy` returns 3 objects so it should be `_, ps, rs = ...`

rampwf/score_types/detection.py
```diff
@@ -652,6 +647,79 @@ def precision_recall_curve(y_true, y_pred, conf_thresholds, iou_threshold=0.5):
     return np.array(ps), np.array(rs)
 
 
+def _add_id(y):
+    """Helper function to flatten and add id column to list of lists"""
+    y_new = []
```
aboucaud: Mention here that this method removes empty lists instead of indexing them.

rampwf/score_types/detection.py
```diff
@@ -652,6 +647,79 @@ def precision_recall_curve(y_true, y_pred, conf_thresholds, iou_threshold=0.5):
...
+    y_pred
+    iou_threshold
+
+    Returns
+    -------
+
+    """
+    y_pred2 = _add_id(y_pred)
+    y_pred2_sorted = y_pred2[np.argsort(y_pred2[:, 1])[::-1], :]
+
```
aboucaud: Add a comment like `# Sorting predicted objects by decreasing confidence`

rampwf/score_types/detection.py
```diff
@@ -652,6 +647,79 @@ def precision_recall_curve(y_true, y_pred, conf_thresholds, iou_threshold=0.5):
...
+    """
+    Calculate precision and recall incrementally based on the predictions
+    sorted by confidence (not calculating an exact optimal match for each
+    confidence level).
+
+    Parameters
+    ----------
+    y_true
+    y_pred
+    iou_threshold
```
aboucaud: Mention here the expected input for y_pred, since its different than y_true 
=> `(conf, row, col, rad)`

rampwf/score_types/detection.py
```diff
@@ -751,6 +819,28 @@ def average_precision_interpolated(ps, rs):
...
 
 
+def average_precision_exact(ps, rs):
+    # from https://github.com/amdegroot/ssd.pytorch/blob/ce4c994db0ee11f82aabb4fdb3499dc970156db5/eval.py#L182-L213
+
+    # correct AP calculation
+    # first append sentinel values at the end
+    mrec = np.concatenate(([0.], rs, [1.]))
+    mpre = np.concatenate(([0.], ps, [0.]))
+
```
aboucaud: Sure this is not a `[1.]` after `ps, ` ?

rampwf/score_types/detection.py
jorisvandenbossche: yep

rampwf/score_types/detection.py
jorisvandenbossche: done

rampwf/score_types/detection.py
jorisvandenbossche: done

rampwf/score_types/detection.py
jorisvandenbossche: Hmm, good question. I copied it from https://github.com/amdegroot/ssd.pytorch/blob/ce4c994db0ee11f82aabb4fdb3499dc970156db5/eval.py#L182-L213, which of course does not mean it is correct.

Intuitively indeed it should be different for precision and recall

jorisvandenbossche: I suppose that because a few lines further, you take the cumulative max along the precision curve, it does not matter (or to say, the 0 is replace by the highest obtained precision, which is not necessarily 1 if the first prediction (the one with the highest confidence) is not a good one)

jorisvandenbossche: same seems to be done by google in tensorflow: https://github.com/tensorflow/models/blob/fb12693f2420cc687d0ca8c640c241610b2b036b/research/object_detection/utils/metrics.py#L110-L120

aboucaud: > I suppose that because a few lines further, you take the cumulative max along the precision curve, it does not matter (or to say, the 0 is replace by the highest obtained precision, which is not necessarily 1 if the first prediction (the one with the highest confidence) is not a good one)

👍  sounds like a good reason why

PR #67, opened 2017-10-17
Add Detection Error score 

jorisvandenbossche:


jorisvandenbossche:
For me it feels a bit strange to only look at the number of objects, and not at the actual match. As the number does not say anything about the quality of the detection. 

We could also calculate the `n_matches`, and then all of `n_true`, `n_pred` and `n_matches` have to be equal instead of only `n_true` and `n_matches`

PR #69, opened 2017-10-19
add the simplified image classifier workflow

mehdidc:


rampwf/workflows/simplified_image_classifier.py
```diff
@@ -0,0 +1,161 @@
...
+    as input an instance of `ImageLoader`.
+
+    Parameters
+    ==========
+
+    n_classes : int
+        Total number of classes.
+
+    """
+    def __init__(self, n_classes, workflow_element_names=['batch_classifier']):
```
jorisvandenbossche: batch_classifier is a bit a strange name if there is no batch anymore

mehdidc: Indeed, maybe we can just use image_classifier.

jorisvandenbossche: I would do that, then with eg a ImageClassifier class the participants need to implement

rampwf/workflows/simplified_image_classifier.py
```diff
@@ -0,0 +1,164 @@
...
+
+    n_classes : int
+        Total number of classes. 
+    """
+
+    def __init__(self, X_array, y_array, folder, n_classes):
+        self.X_array = X_array
+        self.y_array = y_array
+        self.folder = folder
+        self.n_classes = n_classes
```
jorisvandenbossche: this one is not used?

jorisvandenbossche:
In the example in MNIST_simplified, the full dataset is loaded into memory in the actual submission. Is that thought to be the typical workflow? 
Because if that is the case, I would just pass the data to the underlying ImageClassifier instead of the ImageLoader object, which would make the interface much simpler for the participant IMO.

kegl:
No, it's not the typical way this will be used. MNIST is a small set of tiny images. We typically load 16-32-64 jpegs at a time in the submissions. The reason of this workflow is to make accessing the data more flexible, without exposing the physical storage (so e.g. they can open the test images in fit). The typical flexibility this allows them to do is to load the images in random order and write their own generators to be sent to keras.

rampwf/workflows/simplified_image_classifier.py
mehdidc: Indeed, it is not used, but I thought that it coud be needed by users submissions to know the number of classes of the task, in order to to initialize the model.

rampwf/workflows/simplified_image_classifier.py
```diff
@@ -83,6 +84,16 @@ def test_submission(self, trained_model, folder_X_array):
         return y_proba
 
 
+def image_transform(x, transforms):
```
kegl: @mehdidc @jorisvandenbossche I'll need code review here. Remember, the problem is that because of `imp` we can't pickle any user submission, which means they cannot implement parallel loading or image transform. Implementing loading was easy, but for the transform they also can't provide the transform function itself, so we need to implement the transform itself, which they can parameterize. The syntax is a list of dictionaries, each dictionary containing `name` and a list of named parameters. Below is the implementation of a random rotation between `l_angle` and `u_angle`.

jorisvandenbossche:
Shouldn't the transform (rotating images) be something in the user submission? As typically a participant wants to add more of such transform steps?

kegl:
But that's exactly what we can't do (clash between joblib and imp).

Issue #65, opened 2017-10-16
BUG: current SCP calculation failing with ZeroDivisionError


jorisvandenbossche:
Opening here so we don't forget, I just noted that I get an error when calculating SCP for the full test predictions for the ssd model in the starting kit (the pretrained weights):

```
import sys
sys.path.append('./submissions/keras_ssd7/')

from object_detector import ObjectDetector
model = ObjectDetector()
model.fit(X_train, y_train)

X_train, y_train = problem.get_train_data()
X_test, y_test = problem.get_train_data()

y_pred = model.predict(X_test)

from rampwf.score_types.detection import SCP
scp = SCP(shape=(224, 224), minipatch=[56, 168, 56, 168])

scp(y_test, y_pred)
```

gives 

```
/media/joris/DATA/CDS/RAMP/ramp-workflow/rampwf/score_types/detection.py in __call__(self, y_true, y_pred, conf_threshold)
     24              if detected_object[0] > conf_threshold]
     25             for single_detection in y_pred]
---> 26         return self.detection_score(y_true, y_pred_above_confidence)
     27 
     28 

/media/joris/DATA/CDS/RAMP/ramp-workflow/rampwf/score_types/detection.py in detection_score(self, y_true, y_pred)
     82         scps = np.array(
     83             [scp_single(t, p, self.shape, self.minipatch)
---> 84              for t, p in zip(y_true, y_pred)])
     85         return np.sum(scps[:, 0]) / np.sum(scps[:, 1:3])
     86 

/media/joris/DATA/CDS/RAMP/ramp-workflow/rampwf/score_types/detection.py in <listcomp>(.0)
     82         scps = np.array(
     83             [scp_single(t, p, self.shape, self.minipatch)
---> 84              for t, p in zip(y_true, y_pred)])
     85         return np.sum(scps[:, 0]) / np.sum(scps[:, 1:3])
     86 

/media/joris/DATA/CDS/RAMP/ramp-workflow/rampwf/score_types/detection.py in scp_single(y_true, y_pred, shape, minipatch)
    227 
    228     """
--> 229     map_true, map_pred = circle_maps(y_true, y_pred, shape)
    230     if minipatch is not None:
    231         map_true = map_true[

/media/joris/DATA/CDS/RAMP/ramp-workflow/rampwf/score_types/detection.py in circle_maps(y_true, y_pred, shape)
    826     for circle in y_pred:
    827         map_pred = project_circle(
[...]

PR #70, opened 2017-10-19
 Restructure detection scores into multiple files 

jorisvandenbossche:


jorisvandenbossche:
@aboucaud ideas? (now it is maybe a bit too much files)

aboucaud:
Many files indeed, but it was needed IMO.

aboucaud:
Fix #65

PR #78, opened 2017-10-26
Fix NaNs in --quick-test scores

aboucaud:


aboucaud:
@jorisvandenbossche

rampwf/utils/testing.py
```diff
@@ -164,23 +165,24 @@ def assert_submission(ramp_kit_dir='.', ramp_data_dir='.',
...
-    stds = test_scoress.std(axis=0)
+def _print_result(scores, score_types, step):
+    means = scores.mean(axis=0)
+    stds = scores.std(axis=0)
     for mean, std, score_type in zip(means, stds, score_types):
-        print('test {} = {} ± {}'.format(
-            score_type.name, round(mean, score_type.precision),
-            round(std, score_type.precision + 1)))
+        # If std is a NaN
+        if std != std:
```
jorisvandenbossche: Is this needed? 
It feels a bit a way to hide bugs, as the std can never really be NaN when using actual data?

jorisvandenbossche: (ah no, it is NaN if your scores are exactly the same, eg all 0 because of using a dummy model)

rampwf/utils/testing.py
```diff
@@ -164,23 +165,24 @@ def assert_submission(ramp_kit_dir='.', ramp_data_dir='.',
...
+            result = '{step} {name} = {val}'.format(
+                step=step, name=score_type.name, val=mean)
+        else:
+            result = '{step} {name} = {val} ± {std}'.format(
+                step=step,
+                name=score_type.name,
+                val=round(mean, score_type.precision),
+                std=round(std, score_type.precision + 1))
```
jorisvandenbossche: why the plus one?

rampwf/utils/testing.py
```diff
@@ -164,23 +165,24 @@ def assert_submission(ramp_kit_dir='.', ramp_data_dir='.',
...
 
-    means = train_valid_scoress.mean(axis=0)
-    stds = train_valid_scoress.std(axis=0)
-    for mean, std, score_type in zip(means, stds, score_types):
-        print('valid {} = {} ± {}'.format(
-            score_type.name, round(mean, score_type.precision),
-            round(std, score_type.precision + 1)))
+    _print_result(train_train_scoress, score_types, 'train')
+    _print_result(train_valid_scoress, score_types, 'valid')
+    _print_result(test_scoress, score_types, 'test')
```
jorisvandenbossche: Nice clean-up!

rampwf/score_types/detection/precision_recall.py
```diff
@@ -189,7 +195,12 @@ def mad_center(y_true, y_pred, matches=None, iou_threshold=0.5,
     loc_true, loc_pred = _locate_matches(
         y_true, y_pred, matches, iou_threshold=iou_threshold)
 
-    d = np.sqrt((loc_pred[:, 0] - loc_true[:, 0]) ** 2 + (
-        loc_pred[:, 1] - loc_true[:, 1]) ** 2)
+    x_true, y_true, rad_true = loc_true.T
+    x_pred, y_pred, _ = loc_pred.T
+
+    if not x_true:
+        return np.inf
```
jorisvandenbossche: I am wondering a bit, is this better than NaN ?

rampwf/score_types/detection/precision_recall.py
```diff
@@ -155,7 +155,13 @@ def mad_radius(y_true, y_pred, matches=None, iou_threshold=0.5,
     loc_true, loc_pred = _locate_matches(
         y_true, y_pred, matches, iou_threshold=iou_threshold)
 
-    return np.abs((loc_pred[:, 2] - loc_true[:, 2]) / loc_true[:, 2]).mean()
+    _, _, rad_true = loc_true.T
+    _, _, rad_pred = loc_pred.T
```
jorisvandenbossche: Since this are actually arrays, instead of unpacking I would just index (`rad_true = loc_true[:, 2]`)

(but good clean-up to give it names!)

jorisvandenbossche:
There is another case we saw NaNs, what is reported in https://github.com/ramp-kits/mars_craters/issues/43, but that one I cannot reproduce (since I don't really know what is going on there)

rampwf/utils/testing.py
aboucaud: I did not change a thing of what was there.
I suppose it is to have 'precision' digits after the coma.

rampwf/score_types/detection/precision_recall.py
aboucaud: I just wanted to conform to the definition of the score and its 'maximum value'.

rampwf/utils/testing.py
jorisvandenbossche: Ah, I see, it is how it was before. 
I would say, as you are changing, clean it up to remove the '+1'. But I suppose the original author had a reason for it ;-)

(I personally think a standard deviation with such a high precision does not make much sense)

rampwf/score_types/detection/precision_recall.py
aboucaud: Does it create a view instead of a copy ?

jorisvandenbossche: I think both are views (the slicing certainly)

rampwf/score_types/detection/precision_recall.py
jorisvandenbossche: OK, but starting from that definition: [0 - np.inf] is in principle the range of *valid* scores (the lower the bigger the difference in radius, you go towards infinity), while here there is no actual value to calculate. It is just that this score is not defined when there are no matches. And not defined is not necessarily the same as np.inf

But anyhow, doesn't really matter, as both mean there is something really wrong with the model :-)

kegl:
You could use math.isnan, it's better, no?

aboucaud:
Well it's common in Python to use the trick `a != a` for NaNs

aboucaud:
Good to merge I guess

PR #80, opened 2017-10-27
Add mars_craters to the tested kits

jorisvandenbossche:


aboucaud:
Still left with an IOError when executing the notebook.
```python
IOError: [Errno 2] No such file or directory: 'data/data_train.npy'
```

jorisvandenbossche:
> @jorisvandenbossche I see you have kept the "_mini" suffix in the files. Shall we not change that and let it overwrite with the download ?

That what I first did, but:

- the download data checks for the files already existing, and then does nothing. I think it is sensible to keep this behaviour.
- the content of the files get changed, so once you downloaded the full data, git constantly wants to commit those changes, which we don't want (unless you can ignore files that are actually added?)

jorisvandenbossche:
For the error, that is just because the notebook gets executed, and the data are not available, so that is not possible. I don't think we should execute the notebooks here in the testing (we already test the actual submission, which is more related to the ramp-workflow code)

jorisvandenbossche:
I was doing this: https://github.com/paris-saclay-cds/ramp-workflow/pull/81 for that notebook problem.

But I now see that actually only the first notebook gets executed, so maybe we can just switch order and keep the first one executed.

aboucaud:
I see. Nice initiative.

PR #81, opened 2017-10-27
Do not execute ramp-kits notebooks on testing

jorisvandenbossche:


jorisvandenbossche:
Will be fixed in https://github.com/paris-saclay-cds/ramp-workflow/pull/84 by just doing the 'execute' only on titanic

PR #82, opened 2017-10-27
* Add opencv/pillow/skimage.read backend to ImageLoader

vfdev-5:
Tested with 
```
for b in ['opencv', 'pillow', 'skimage']:
    print("Test ", b)
    img_loader.setup_backend(b)
    for i in range(100):
        img_loader.load(i)

    _ = img_loader.parallel_load(list(range(100)))    
```

### DIFF
--- a/rampwf/workflows/simplified_image_classifier.py
+++ b/rampwf/workflows/simplified_image_classifier.py
@@ -3,6 +3,10 @@ import os
 import imp
 import numpy as np
 
+from PIL import Image
+import cv2
+from skimage.io import imread as skimage_imread
+
 
 class SimplifiedImageClassifier(object):
     """
@@ -95,6 +99,27 @@ def _image_transform(x, transforms):
     return x
 
 
+def _imread_opencv(filename):
+    img = cv2.imread(filename)
+    if img is not None:
+        # RGBA -> RGB
+        if img.shape[2] == 4:
+            img = img[:, :, 0:3]
+        return cv2.cvtColor(img, cv2.COLOR_BGR2RGB)
+    else:
+        # It can happen that Opencv fails to open a file
+        # Open it with skimage
+        img = skimage_imread(filename)
+        if img.shape[2] == 4:
+            img = img[:, :, 0:3]
+        return img
+
+
+def _imread_pillow(filename):
+    img = Image.open(filename)
+    return np.asarray(img)
+
+
 class ImageLoader(object):
     """
     Load and image and optionally its label.
@@ -126,12 +151,23 @@ class ImageLoader(object):
         Total number of classes.
     """
 
-    def __init__(self, X_array, y_array, folder, n_classes):
+    def __init__(self, X_array, y_array, folder, n_classes, backend='skimage'):
         self.X_array = X_array
         self.y_array = y_array
         self.folder = folder
         self.n_classes = n_classes
         self.nb_examples = len(X_array)
+        self.setup_backend(backend)
+
+    def setup_backend(self, backend):
+        assert backend in ['opencv', 'pillow', 'skimage']
+        self.backend = backend
+        if backend == 'opencv':
+            self._imread = _imread_opencv
+        elif backend == 'pillow':
+            self._imread = _imread_pillow
+        elif backend == 'skimage':
+            self._imread = skimage_imread
 
     def load(self, index):
         """
@@ -158,14 +194,13 @@ class ImageLoader(object):
         a tuple (x, y).
         At test time, `y_array` is `None`, and `load` returns `x`.
         """
-        from skimage.io import imread
-
         if index < 0 or index >= self.nb_examples:
             raise IndexError("list index out of range")
 
         x = self.X_array[index]
         filename = os.path.join(self.folder, '{}'.format(x))
-        x = imread(filename)
+        x = self._imread(filename)
+
         if self.y_array is not None:
             y = self.y_array[index]
             return x, y
@@ -197,7 +232,6 @@ class ImageLoader(object):
         a tuple (x, y).
         At test time, `y_array` is `None`, and `load` returns `x`.
         """
-        from skimage.io import imread
         from joblib import delayed, Parallel, cpu_count
 
         for index in indexes:
@@ -208,7 +242,7 @@ class ImageLoader(object):
             os.path.join(self.folder, '{}'.format(self.X_array[index]))
             for index in indexes]
         xs = Parallel(n_jobs=n_jobs, backend='threading')(
-            delayed(imread)(filename) for filename in filenames)
+            delayed(self._imread)(filename) for filename in filenames)
 
         if transforms is not None:
             from functools import partial
--- a/rampwf/workflows/simplified_image_classifier.py
+++ b/rampwf/workflows/simplified_image_classifier.py
@@ -3,10 +3,6 @@ import os
 import imp
 import numpy as np
 
-from PIL import Image
-import cv2
-from skimage.io import imread as skimage_imread
-
 
 class SimplifiedImageClassifier(object):
     """
@@ -100,6 +96,8 @@ def _image_transform(x, transforms):
 
 
 def _imread_opencv(filename):
+    import cv2
+    from skimage.io import imread as skimage_imread
     img = cv2.imread(filename)
     if img is not None:
         # RGBA -> RGB
@@ -116,6 +114,7 @@ def _imread_opencv(filename):
 
 
 def _imread_pillow(filename):
+    from PIL import Image
     img = Image.open(filename)
     return np.asarray(img)
 
@@ -167,6 +166,7 @@ class ImageLoader(object):
         elif backend == 'pillow':
             self._imread = _imread_pillow
         elif backend == 'skimage':
+            from skimage.io import imread as skimage_imread
             self._imread = skimage_imread
 
     def load(self, index):


aboucaud:
Hi, we would like to keep the dependency list of ramp-workflow minimal, so please consider moving the imports within the methods, so that no `ImportError` is raised during testing.

On the other hand, feel free to create a PR to add these requirements to the `requirements.txt` file of the starting kits using these methods.

glemaitre:
@vfdev-5 could you emphasize the need of having different backend. The backend are high-level backend and it should not matter to use one or another. I would prefer to always use skimage which can use imageio plugin which can be easily extended for any new format.

So if you could motivate this choice that would be useful.

vfdev-5:
@glemaitre, this PR is related to a slack discussion in the context of pollenating insects 3 simplified challenge. Purpose of different backends is to choose a method to load image:
- vanilla skimage
- a bit faster opencv (however can fail to load rgba and some other images)
- pillow (not completely implemeted) idea is to provide dataflow input to pytorch/torchvision framework

glemaitre:
`imread` is using PIL as the first plugin by default: http://scikit-image.org/docs/dev/api/skimage.io.html#imread
So I don't think that there is a need for PIL backend, isn't it?

If opencv is only a bit faster but it is failing, I would not consider it if opening file is not the bottleneck.

I am not really familiar with the deep net framework, but the file opening with the possible data augmentation or other processing are not the bottleneck, isn't it?

ping @kegl could have some insight?

@vfdev-5 I could not find back the discussion in slack. Which channel was it and what date?

kegl:
Image loading does seem to be a bottleneck sometimes. But I'm also not an expert. I was waiting for @mehdidc to do some tests. The discussion is in the ramp slack https://ramp-studio.slack.com. @vfdev-5 can you ping @glemaitre from inside the slack, I can't find the thread (slack seems to be down BTW).

vfdev-5:
Good to know that skimage uses pillow as default image reader. In that case pillow backend option can directly return PIL.Image for torchvision data processing. The last step is to use then pillow-simd instead of standard pillow.

On opencv option, in case of pollenating insects 3 challenge, it fails on rgba images, but still gives a gain, that I should measure the value.

glemaitre:
For an image of (3072, 3072, 3), I get the following benchmark:

```python
import cv2
cv2.imread('Astronaut-EVA.jpg')
232 ms ± 9.57 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

```python
import numpy as np
from PIL import Image
%timeit np.array(Image.open('Astronaut-EVA.jpg'))
255 ms ± 2.32 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

```python
import skimage.io
%timeit skimage.io.imread('Astronaut-EVA.jpg')
290 ms ± 2.93 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

glemaitre:
So the advantage of using PIL is to use directly the torchvision from pytorch:
http://pytorch.org/docs/master/torchvision/transforms.html#transforms-on-pil-image

Therefore I would be inclined to always use PIL directly and either return `PIL.image` or convert it to `np.array` depending of a keyword. If I am not wrong `PIL.Image.open()` should be a lazy reading and will only happen when the data are needed. If really the data need to be loaded, you can use `PIL.image.load()`.

vfdev-5:
Here are my results on quarter of the training dataset:

```b = 'skimage'
n_epochs = 5
n_samples = len(img_loader)/4

img_loader.setup_backend(b)
mean_time = 0

for e in range(n_epochs):
    print("Epoch : %i / %i, n_samples=%i" % (e + 1, n_epochs, n_samples))
    start = time()
    for i in range(n_samples):
        img_loader.load(i)
        if i % 1000 == 0:
            print(i, end=' . ')
    end = time()
    elapsed = end - start
    mean_time += elapsed
    print("Epoch time: {}".format(elapsed))
print("Mean epoch time: {}".format(mean_time / n_epochs))

Epoch : 1 / 5, n_samples=11670
0 . 1000 . 2000 . 3000 . 4000 . 5000 . 6000 . 7000 . 8000 . 9000 . 10000 . 11000 . Epoch time: 148.139643908
Epoch : 2 / 5, n_samples=11670
0 . 1000 . 2000 . 3000 . 4000 . 5000 . 6000 . 7000 . 8000 . 9000 . 10000 . 11000 . Epoch time: 148.038290024
Epoch : 3 / 5, n_samples=11670
0 . 1000 . 2000 . 3000 . 4000 . 5000 . 6000 . 7000 . 8000 . 9000 . 10000 . 11000 . Epoch time: 148.08396697
Epoch : 4 / 5, n_samples=11670
0 . 1000 . 2000 . 3000 . 4000 . 5000 . 6000 . 7000 . 8000 . 9000 . 10000 . 11000 . Epoch time: 147.941848993
Epoch : 5 / 5, n_samples=11670
0 . 1000 . 2000 . 3000 . 4000 . 5000 . 6000 . 7000 . 8000 . 9000 . 10000 . 11000 . Epoch time: 148.144033909
Mean epoch time: 148.069556761
```

and

```b = 'opencv'
n_epochs = 5
n_samples = len(img_loader)/4

img_loader.setup_backend(b)
mean_time = 0

for e in range(n_epochs):
    print("Epoch : %i / %i, n_samples=%i" % (e + 1, n_epochs, n_samples))
    start = time()
    for i in range(n_samples):
        img_loader.load(i)
        if i % 1000 == 0:
            print(i, end=' . ')
    end = time()
    elapsed = end - start
    mean_time += elapsed
    print("Epoch time: {}".format(elapsed))
print("Mean epoch time: {}".format(mean_time / n_epochs))

Epoch : 1 / 5, n_samples=11670
0 . 1000 . 2000 . 3000 . 4000 . 5000 . 6000 . 7000 . 8000 . 9000 . 10000 . 11000 . Epoch time: 123.776700974
Epoch : 2 / 5, n_samples=11670
0 . 1000 . 2000 . 3000 . 4000 . 5000 . 6000 . 7000 . 8000 . 9000 . 10000 . 11000 . Epoch time: 121.414109945
Epoch : 3 / 5, n_samples=11670
0 . 1000 . 2000 . 3000 . 4000 . 5000 . 6000 . 7000 . 8000 . 9000 . 10000 . 11000 . Epoch time: 123.362024069
Epoch : 4 / 5, n_samples=11670
0 . 1000 . 2000 . 3000 . 4000 . 5000 . 6000 . 7000 . 8000 . 9000 . 10000 . 11000 . Epoch time: 122.042551041
Epoch : 5 / 5, n_samples=11670
0 . 1000 . 2000 . 3000 . 4000 . 5000 . 6000 . 7000 . 8000 . 9000 . 10000 . 11000 . Epoch time: 121.966143131
Mean epoch time: 122.512305832```

If data loading is the bottleneck in the pipeline, opencv can give a gain during the training phase.

**Comment by glemaitre** [comment]
@vfdev-5 Can you report pillow as well since that it should be some skimage without some call to warped function?

**Comment by glemaitre** [comment]
IMHO, I would remove the skimage backend (it could limit for some image format) and use pillow instead. I would also make no conversion to numpy array since that the transformer could take advantage of the lazy reading (it means as well that the parallel reading is useless because the overhead will be too important).

We can still think about putting opencv (I am personally not a big fan:)) but I will let this to other reviewers to give some pros/cons argument.

@vfdev-5 We would need to have some unit tests as well.

**Comment by vfdev-5** [comment]
Results of `pillow` with `np.asarray` (standard, not pillow-simd) give:
```
b = 'pillow'
n_epochs = 5
n_samples = len(img_loader)/4

img_loader.setup_backend(b)
mean_time = 0

for e in range(n_epochs):
    print("Epoch : %i / %i, n_samples=%i" % (e + 1, n_epochs, n_samples))
    start = time()
    for i in range(n_samples):
        img_loader.load(i)
        if i % 1000 == 0:
            print(i, end=' . ')
    end = time()
    elapsed = end - start
    mean_time += elapsed
    print("Epoch time: {}".format(elapsed))
print("Mean epoch time: {}".format(mean_time / n_epochs))

Epoch : 1 / 5, n_samples=11670
0 . 1000 . 2000 . 3000 . 4000 . 5000 . 6000 . 7000 . 8000 . 9000 . 10000 . 11000 . Epoch time: 128.705247164
Epoch : 2 / 5, n_samples=11670
0 . 1000 . 2000 . 3000 . 4000 . 5000 . 6000 . 7000 . 8000 . 9000 . 10000 . 11000 . Epoch time: 128.782455921
Epoch : 3 / 5, n_samples=11670
0 . 1000 . 2000 . 3000 . 4000 . 5000 . 6000 . 7000 . 8000 . 9000 . 10000 . 11000 . Epoch time: 128.739259958
Epoch : 4 / 5, n_samples=11670
0 . 1000 . 2000 . 3000 . 4000 . 5000 . 6000 . 7000 . 8000 . 9000 . 10000 . 11000 . Epoch time: 128.678661823
Epoch : 5 / 5, n_samples=11670
0 . 1000 . 2000 . 3000 . 4000 . 5000 . 6000 . 7000 . 8000 . 9000 . 10000 . 11000 . Epoch time: 128.828619003
Mean epoch time: 128.746848774
```
I'm more fan of opencv than pillow and I would leave the possibility to work with `np.array` and perform all transformations with opencv. Here is a small benchmark between opencv and standard pillow: [link](https://www.kaggle.com/vfdev5/pil-vs-opencv)

PR #84, opened 2017-10-28
porting cv bagging from ramp-board to ramp-workflow

kegl:
Not much code but will need careful review. Make it more pythonesque.

### DIFF
--- a/rampwf/utils/combine.py
+++ b/rampwf/utils/combine.py
@@ -1,0 +1,51 @@
+import numpy as np
+
+
+def get_score_cv_bags(Predictions, score_type, predictions_list, ground_truths,
+                      test_is_list=None):
+    """
+    Compute the bagged scores of the predictions in predictions_list.
+
+    test_is_list (list of list of integer indexes) controls which points
+    in which fold participate in the combination. We return the
+    combined predictions and a list of scores, where each element i is the
+    score of the combination of the first i+1 folds.
+
+    Parameters
+    ----------
+    Predictions : a class implementing BasePrediction signature
+        Needed for the type of y_comb
+    score_type : instance implementing BaseScoreType signature
+    predictions_list : list of instances of Predictions
+    ground_truths : instance of Predictions
+    test_is_list : list of list of integers
+        Indices of points that should be bagged in each prediction. If None,
+        the full prediction vectors will be bagged.
+    Returns
+    -------
+    combined_predictions : instance of Predictions
+    score_cv_bags : list of scores (typically floats)
+    """
+    if test_is_list is None:  # we combine the full list
+        test_is_list = [range(len(predictions.y_pred))
+                        for predictions in predictions_list]
+
+    y_comb = np.array(
+        [Predictions(n_samples=len(ground_truths.y_pred))
+         for _ in predictions_list])
+    score_cv_bags = []
+    for i, test_is in enumerate(test_is_list):
+        # setting valid fold indexes of points to be combined
+        y_comb[i].set_valid_in_train(predictions_list[i], test_is)
+        # combine first i folds
+        combined_predictions = Predictions.combine(y_comb[:i + 1])
+        # get indexes of points with at least one prediction in
+        # combined_predictions
+        valid_indexes = combined_predictions.valid_indexes
+        # score the combined predictions
+        score_cv_bags.append(score_type.score_function(
+            ground_truths, combined_predictions, valid_indexes))
+        # Alex' old suggestion: maybe use masked arrays rather than passing
+        # valid_indexes
+    # TODO: will crash if len(test_is_list) == 0
+    return combined_predictions, score_cv_bags
--- a/rampwf/utils/tests/test_testing.py
+++ b/rampwf/utils/tests/test_testing.py
@@ -13,12 +13,17 @@ def test_submission_all_kits():
     try:
         for kit in RAMP_KITS_AVAILABLE:
             kit_dir = fetch_ramp_kit(kit, ramp_kits_home=tmp_dir)
-            assert_submission(
-                ramp_kit_dir=kit_dir, ramp_data_dir=kit_dir,
-                submission='starting_kit', is_pickle=True, save_y_preds=True)
-            # testing assert_notebook on titanic
+            # testing assert_notebook and optional switches on titanic
             if kit == 'titanic':
                 assert_notebook(ramp_kit_dir=kit_dir)
+                assert_submission(
+                    ramp_kit_dir=kit_dir, ramp_data_dir=kit_dir,
+                    submission='starting_kit', is_pickle=True,
+                    save_y_preds=True, retrain=True)
+            else:
+                assert_submission(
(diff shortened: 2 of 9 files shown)


agramfort:
@kegl can help me understand the problem you're solving here?

kegl:
Increasing coverage :)

agramfort:
you have my full support. But why is this increasing it?

kegl:
1. Porting cv-bagging into the test.
2. Adding also classical CV setup (model is retrained on full training set).
3. Letting the users (optionally) pickle their trained models locally.
4. Letting the users (optionally) save their y_preds (e.g. if they want to submit test predictions to Kaggle). The default is as simple `np.savetxt`. If it doesn't work (e.g. mars craters), we catch the error and tell it to the users that they should implement a `save_y_pred` method in `problem.py`. This also allows them to follow the format they want, e.g., in `kaggle_seguro` we need to re-read `X_test` to get the ids outputted into the submission `y_test_pred`.

The higher level reason of all of this is to continue porting functionalities of `ramp-board` into `ramp-workflow` so the starting kits can be tested properly. E.g. `BasePrediction.combine` was not covered before, causing `ramp-board` crashing even when starting kit passed. But this is only a side effect, the goal is to give these functionalities to the user. The last such thing will be model combination, which we will be able to implement now that we can save the predictions.

kegl:
One more thing which is getting easier now is to write a script that optimizes/makes a report on CV parameters (number of folds and train/test proportions).

kegl:
@jorisvandenbossche @aboucaud I'll need help here. Both travis https://travis-ci.org/paris-saclay-cds/ramp-workflow/jobs/294574847 and local pytest crashes when it doesn't find the file `mars_craters/data/sample_submission.csv`. It's weird that it is looking for it. It's a file that `save_y_pred` in `kaggle_seguro/problem.py` requires, but what does this have to do with mars craters? 

It's like in this line https://github.com/paris-saclay-cds/ramp-workflow/blob/cv_bagging/rampwf/utils/testing.py#L117 when it doesn't find `save_y_pred` in `mars_craters/problem.py`, instead of raising `AttributeError`, it falls back to the previous call's `kaggle_seguro/problem.py`, and crashes there (when can't find `data/sample_submission.csv`). Weird.

kegl:
Indeed, if I take off `kaggle_seguro` from the top of the list here, mars craters (and the whole test) runs with no problem. https://github.com/paris-saclay-cds/ramp-workflow/blob/cv_bagging/rampwf/kits/ramp_kit.py#L11

aboucaud:
From what I could understand, travis is downloading the mars crater data *before* the testing of the `kaggle_seguro` is complete. Therefore the `data_path` variable set to '.' refers to the `mars_crater` directory and not the `kaggle_seguro` one, hence the trouble we're having. 

We need to add some coordination.

kegl:
It happens also locally, running pytest, but I guess it's the same problem. I could get around it by catching `Exception` instead of `AttributeError`, but it's somewhat unsatisfying. If what you're saying is true, we may have other sync problems. The good news is that this is strictly a testing issue since in no other places we will loop through different starting kits.

agramfort:
you have no async code. How can this happen?

kegl:
OK, I found the problem, it was the same that made imp-ed modules unpickleable: we forgot to name the modules.

kegl:
I think I'm finished. Please @aboucaud @glemaitre @jorisvandenbossche review the code. Also add comments where you think I should comment more, in case some of the additions are unclear. I wouldn't be against some modularization, `assert_submission` is pretty long now.

jorisvandenbossche:
@kegl I like the idea of coloring the output of assert_test_submission to make the overview of the output better, but I think we need a solution that is also compatible with Windows (`colored` mentions it is only for linux/unix, and @glemaitre also quickly tested it on windows and it does not seem to work). 
There are terminal coloring libraries that can do that I think, but I would personally leave that for a separate PR.

jorisvandenbossche:
Small miscommunication with @glemaitre: it does not work on Windows, but also does not output any garbage (it just ignores it). So then might be fine to keep, although it would still be nice to have something that works cross-platform

requirements.txt
```diff
@@ -2,4 +2,5 @@ numpy>=1.13
 scipy>=0.19
 pandas>=0.19.2
 scikit-learn>=0.18
-cloudpickle
\ No newline at end of file
+cloudpickle
```
jorisvandenbossche: if you add something here, you should also add it in the `install_requires` list in setup.py (that is the one which is used when pip installing ramp-workflow)

rampwf/utils/tests/test_testing.py
```diff
@@ -11,15 +11,14 @@
...
     try:
-        test_notebook = True
         for kit in RAMP_KITS_AVAILABLE:
             kit_dir = fetch_ramp_kit(kit, ramp_kits_home=tmp_dir)
             assert_submission(
                 ramp_kit_dir=kit_dir, ramp_data_dir=kit_dir,
-                submission='starting_kit')
-            # testing assert_notebook on first kit
-            if test_notebook:
+                submission='starting_kit', is_pickle=True, save_y_preds=True)
```
jorisvandenbossche: retrain=True as well?

rampwf/utils/combine.py
```diff
@@ -0,0 +1,51 @@
...
+        test_is_list = [range(len(predictions.y_pred))
+                        for predictions in predictions_list]
+
+    y_comb = np.array(
+        [Predictions(n_samples=len(ground_truths.y_pred))
+         for _ in predictions_list])
+    score_cv_bags = []
+    for i, test_is in enumerate(test_is_list):
+        # setting valid fold indexes of points to be combined
+        y_comb[i].set_valid_in_train(predictions_list[i], test_is)
```
jorisvandenbossche: Does this work if y_pred are not just numbers? (or should then the Predictions override the base `set_valid_in_train` ?)

agramfort:
what's the meaning of colors?

rampwf/utils/combine.py
kegl: Yes, the latter.

rampwf/utils/tests/test_testing.py
kegl: I actually rethought this a bit. For titanic we test `assert_notebook` and the optional switches (to test the script itself). For the others, we don't need the switches (remember the kits' role here is to test functionalities of rampwf; then in the travis of every kit, we should test the different ways of calling the script for that particular kit).

requirements.txt
kegl: Was done by @aboucaud.

kegl:
I agree that we need to cut up `assert_submission`, will work on this.

I made the warning disappear, a bit hacky, see next commit.

Yes, I went overboard with colors. I like that the yellow stands out and shows me where the run started when I scroll back. Then green/blue/red is train/valid/test, and bright means the official score. My main purpose in the beginning was to make the official test scores stand out (bright red). Let's keep this open for now, see how it feels after getting used to it.

PR #85, opened 2017-10-30
[WIP] Colored print

aboucaud:


jorisvandenbossche:
Can you provide a bit of context? How does this relate to https://github.com/paris-saclay-cds/ramp-workflow/pull/84 ?

aboucaud:
This PR follows on #84 and 
- add a fallback
- avoids adding a dependency in the code

jorisvandenbossche:
I think it might be rather easy to use colorama on windows, to make this work there as well (https://pypi.python.org/pypi/colorama). Might be it just needs a `colorama.init()` (but didn't test).

It is a dependency of IPython on windows, so probably most have it installed on that platform (so then not needed to make it a hard dependency)

aboucaud:
I'll close this PR and wait for #84 to be merged before doing something.

PR #86, opened 2017-10-31
[WIP] README refactoring after wiki creation

aboucaud:
IMO a huge README where scrolling gets you lost is not very attractive.
Also a lot of it were specific examples from titanic.

Please comment..

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,10 +35,10 @@ $ pip install git+https://github.com/paris-saclay-cds/ramp-workflow.git
 This will set up some command line scripts like `ramp_test_submission`.
 We suggest to use a dedicated virtual environment if you are familiar with it.
 
-2. Pick up a starting-kit on <https://github.com/ramp-kits>
+2. Pick a starting-kit on <https://github.com/ramp-kits>
 
-Clone it locally, and fire the starting kit notebook.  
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Instead of moving it to the wiki, what about putting those files in a /docs/ directory ?

aboucaud:
Well, the wiki is kinda "made for this".  
Otherwise we simply create a documentation webpage..  
But I like the wiki visual.

kegl:
Aren't those two things (wiki and docs) complementary? I like the wiki as is, it's clear and short, we can start marketing it communicate. In the docs I would put more detailed docs on the code, no?

kegl:
LGTM

jorisvandenbossche:
> Well, the wiki is kinda "made for this".
Otherwise we simply create a documentation webpage..

Well, typically I expect to find things as "how to install this", "how to build my own workflow", .. in the documentation. Whether this docs is hosted as a wiki or as a documentation webpage does not matter that much to me as a user, as long it is clear. Therefore I would certainly avoid having both wiki with the current content ànd documentation page. 
As a developer, I personally prefer a /docs/ subdir for user oriented documentation, as this is then version controlled together with the rest of the package.

If it is because you like more the markdown, we can also just use markdown in a /docs/ subdir.

PR #87, opened 2017-10-31
[MRG + 1] Use versioneer for version management

jorisvandenbossche:
This currently gives:

```
In [1]: import rampwf

In [2]: rampwf.__version__
Out[2]: 'v0.1.0+16.g28f84af'
```


jorisvandenbossche:
Note: all those files are automatically generated and should never be adapted by hand (otherwise it would be a lot of code to maintain that this adds ..)

kegl:
Travis fails
```
Processing /home/travis/build/paris-saclay-cds/ramp-workflow
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-gCmmfo-build/setup.py", line 28, in <module>
        VERSION = version
    NameError: name 'version' is not defined
```
otherwise LGTM.

jorisvandenbossche:
Whoops, fixed that

glemaitre:
LGTM

PR #89, opened 2017-11-01
Pretty print CV scores

rth:
Currently, when `ramp_test_submission` is used, scores are printed on different subsets of the data for each CV fold. These are somewhat difficult to parse, particularly if the number of scores becomes large (thinking `mars_craters`).

For instance, for `california_rainfall` we will get,
```
CV fold 0
        train BSS = -1.489
        valid BSS = -2.388
        test BSS = -1.386
        train BS = 0.244
        valid BS = 0.246
        test BS = 0.242
        train BS Rel = 0.018
        valid BS Rel = 0.014
        test BS Rel = 0.023
        train BS Res = 0.073
        valid BS Res = 0.115
        test BS Res = 0.111
        train AUC = 0.712
        valid AUC = 0.806
        test AUC = 0.773
```

This PR uses pandas to pretty print a table of scores instead. The previous example would thus produce,
```
CV fold 0
        score    AUC     BS  BS Rel  BS Res    BSS
        step
        train  0.712  0.244   0.018   0.073 -1.489
        test   0.773  0.242   0.023   0.111 -1.386
        valid  0.806  0.246   0.014   0.115 -2.388
```

The only point currently not supported is `c_prefix = 'official_'` could someone explain when this is used? I haven't seen its use when running the test suite.

This only changes the score visualization for CV folds, if this PR is accepted, the "Mean CV scores" table could also be refactored to use the same approach.

Please let me know if you have any comments. Thanks.

### DIFF
--- a/rampwf/utils/testing.py
+++ b/rampwf/utils/testing.py
@@ -22,15 +22,17 @@ fg_colors = {
     'train': 'dark_sea_green_3b',
     'valid': 'light_slate_blue',
     'test': 'pink_1',
+    'title': 'gold_3b',
+    'warning': 'grey_46',
 }
 
 
 def _print_title(str):
-    print(stylize(str, fg('gold_3b') + attr('bold')))
+    print(stylize(str, fg(fg_colors['title']) + attr('bold')))
 
 
 def _print_warning(str):
-    print(stylize(str, fg('grey_46')))
+    print(stylize(str, fg(fg_colors['warning'])))
 
 
 def _delete_line_from_file(f_name, line_to_delete):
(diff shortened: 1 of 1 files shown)


kegl:
I like this.

We use `official_` to color the official scores differently (brighter). Official score is the first in the `problem.score_types` list. We also have a color scheme now, it may be less important with your table, but I think it would still add visual cue, can you use them in the table? Can you make it such that the table columns follow the score order in  `problem.score_types`? In the mars_craters kit we have many scores (it's also a good testbed for this), and the table is cut (at 80 characters I think). I would be happier if the table was not cut (I can resize my terminal), but maybe xterm junkies wouldn't like this.

kegl:
There are also some flake issues (travis failed).

rth:
Thanks for the review @kegl !

> We use official_ to color the official scores differently (brighter). Official score is the first in the problem.score_types list. We also have a color scheme now, it may be less important with your table, but I think it would still add visual cue, can you use them in the table?

I added colored lines. This will make the code more fragile though and increases the number of ways it could break. Adding the special color for the first column would be more difficult..

> Can you make it such that the table columns follow the score order in problem.score_types?

Done.

> In the mars_craters kit we have many scores (it's also a good testbed for this), and the table is cut (at 80 characters I think). I would be happier if the table was not cut 

Done.

> There are also some flake issues (travis failed).
> ```
> flake8 rampwf --ignore=F401,E211,E265,W503
> rampwf/utils/testing.py:179:9: E722 do not use bare except'
> rampwf/utils/testing.py:308:17: E126 continuation line over-indented for hanging indent
> ```

Fixed the bare except. For the E126 issue, it's a non-default flake8 error (the current `--ignore` is more restrictive than the [default flake8 options](https://github.com/PyCQA/flake8/blob/52180995046583cacf9d78a4423b82448a47fef1/docs/source/manpage.rst#L59)), that is not enforced by PEP8... 
> (from  [here](https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes))
> In the default configuration, the checks E121, E123, E126, E133, E226, E241, E242, E704 and W503
 are ignored because they are not rules unanimously accepted, and PEP 8 does not enforce them.

Also related to the discussion on the effect of the overly strict flake8 options on the contribution workflow in https://github.com/scikit-learn/scikit-learn/issues/9121 ...

kegl:
Not sure what's your issue with flake there: just lose a tab here 
```
        df_scores = _score_matrix(
            score_types,
            ground_truth=OrderedDict([('train', ground_truth_train_train),
                                      ('valid', ground_truth_train_valid),
                                      ('test', ground_truth_test)]),
            predictions=OrderedDict([('train', predictions_train_train),
                                     ('valid', predictions_train_valid),
                                     ('test', predictions_test)]),
            indent='\t', verbose=True)
```

kegl:
Can you use the brighter color for the first (official) column?

You could also redo the mean scores? The retrain (use --retrain switch) and bagged scores are somewhat different (less scores), but it would be useful to retublate them too if not too complicated.

kegl:
Also, small cosmetics: change the order to train/valid/test please.

kegl:
Also, I think we could delete the line that starts with "step", to make it more concise.

rth:
Thanks for the comments!

> Not sure what's your issue with flake there: just lose a tab here

The problem is that 
```
flake8 rampwf/utils/command_line.py 
```
does not raise any error, while this Travis CI setup does. Since the raised error is not enforced by PEP8, having arbitrary linting requirements is not very friendly towards contributors.

Modified the indent here, just a comment...

> Can you use the brighter color for the first (official) column?

Done

> Also, small cosmetics: change the order to train/valid/test please.

Done

> Also, I think we could delete the line that starts with "step", to make it more concise.

Done

> You could also redo the mean scores? The retrain (use --retrain switch) and bagged scores are somewhat different (less scores), but it would be useful to retublate them too if not too complicated.

Could this be done in some later PR? Since the scores for each CV fold are now contained in a single data frame, potentially one can just concatenate scores for different folds and more easily calculate the mean std etc, but this would require some more refactoring and I was hoping we could keep this PR small..

kegl:
Sorry, I was not clear. The colors were fine, I'd just like the _first_ column (the official score) to be brighter.

rth:
> Sorry, I was not clear. The colors were fine, I'd just like the first column (the official score) to be brighter.

OK, so the idea in this PR was to use pandas to most of the work of rendering tables.

Coloring rows can kind of be done with this approach, but coloring both rows and columns in different colors become more problematic (the only approaches I can think of would be hacky) unless [pandas stlying](http://pandas.pydata.org/pandas-docs/version/0.19.2/style.html) can work in a terminal. If not it would probably be easier and cleaner to write a custom formatting function that supports coloring (which is a bit outside of the scope of this PR). 

Besides, I'm not sure if coloring tables would actually help readability. Below is the table of scores using different color schemes with a black (e.g. Travis CI / Terminal) and a white (e.g. jupyter) backgrounds,

img3

img2

the second column (non-bright colors)  is a bit hard to read with a white background (particularly considering it could be shown on some video-projector with suboptimal lighting conditions during RAMP). To avoid it, one would need different color schemes depending on the background. Then if one wants to to look at e.g. the `sacc` score column for different subsets of the data, I would argue that different colors only hurt readability. Also, I'm not sure what this would look like for color blind people. 

So all this would be some amount of work which is getting a bit far from the original goal of just having a human-readable table. IMO just using whatever formatting is done by pandas in black & white is probably more robust and sufficient for most cases ...

What do you think?

kegl:
I already pushed the fix :). Check my latest commit.

rth:
Sure, that work.. Thanks.

kegl:
OK team @jorisvandenbossche @aboucaud @rth @agramfort @glemaitre it's mostly beautification and some modularization, please review then we'll merge. Travis is failing with a segfault, no idea why, doesn't seem to be related to this PR.

PR #96, opened 2017-11-27
Update import source name to allow caching multiple submissions

jorisvandenbossche:
And at the same time made it into a utility function to avoid repeated code.

This allows to do a `feature_extractor.py` with something like:

```
from joblib import Memory
memory = Memory(cachedir='./cache', verbose=1)


class FeatureExtractor():
    def __init__(self):
        pass

    def fit(self, X_df, y):
        pass

    def transform(self, X_df):
        return _transform(X_df)


@memory.cache
def _transform(X_df):
    ...
    computationaly expensive feature extraction
    ...
    return X_df_transformed
```

The above already works, but then if you have multiple submissions with a different transform method (but same name), they would each time overwrite each other cached results. This patch solves that, and is in general also a good clean-up IMO.

### DIFF
--- a/rampwf/utils/importing.py
+++ b/rampwf/utils/importing.py
@@ -1,3 +1,7 @@
+"""
+Utility to import local files from the filesystem as modules.
+
+"""
 import imp
 import os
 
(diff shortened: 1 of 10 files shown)


glemaitre:
I don't think that the error is related

PR #99, opened 2017-12-23
[WIP] attempt to fix travis

agramfort:
see https://travis-ci.org/paris-saclay-cds/ramp-workflow/jobs/320167139


agramfort:
I cannot replicate locally :(

I would suggest to install netcdf4 from conda rather than pypi to try to debug.

I won't try more here.

jorisvandenbossche:
Should be green now (installed only netcdf in advance with conda)

PR #101, opened 2018-02-04
Catching negative probes in soft accuracy

kegl:


agramfort:
travis is broken but I think it's unrelated.

https://travis-ci.org/paris-saclay-cds/ramp-workflow/jobs/337281928

PR #105, opened 2018-02-05
TST: ensure test passes for python 2

jorisvandenbossche:


agramfort:
does anyone has the time to fix travis?

PR #112, opened 2018-02-06
Suppress Runtime Warnings when importing from starting kits

aboucaud:
Runtime warnings have appeared recently when running the `ramp_test_submission` looking like
```bash
./submissions/starting_kit/feature_extractor.py:2: RuntimeWarning: Parent module '.submissions.starting_kit' not found while handling absolute import
  import numpy as np
```
This is quite annoying to find these warnings in the prompt so I just filtered them in this PR.
We might need to find out why these warnings have appeared..

### DIFF
--- a/rampwf/utils/importing.py
+++ b/rampwf/utils/importing.py
@@ -4,6 +4,7 @@ Utility to import local files from the filesystem as modules.
 """
 import imp
 import os
+import warnings
 
 
 def import_file(module_path, filename):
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
> We might need to find out why these warnings have appeared..

Yes, I think we should do that.

jorisvandenbossche:
Although, looking at the travis log (from before this was pushed), this seems only happening on Python 2, so maybe not really a priority.

(BTW @aboucaud you should use python 3 ! :-))

aboucaud:
Good thing you caught this. 

Yeah I'm using Python2 to set up the starting kit because I want to avoid as much as possible version conflict on the hack day. And astro people tend to stick to python2 for some reason, so..

PR #113, opened 2018-02-06
Avoid redundant loading of training data

aboucaud:
In the `testing.py` file, the `assert_*` methods tend to perform redundant tasks.

In particular, calling all other assert methods within `assert_submission` does load the `problem.py` multiple times and loads the training data twice (once through `assert_data` and once through `assert_cv`). 

This PR aims at clarifying those calls.
The first commit solves the double loading in `assert_submission` but does not touch any other method.

Should we **keep** these assert methods or simply **remove them and inline their content** in `assert_submission` ? @jorisvandenbossche @glemaitre @kegl

### DIFF
--- a/rampwf/utils/testing.py
+++ b/rampwf/utils/testing.py
@@ -532,9 +532,14 @@ def assert_submission(ramp_kit_dir='.', ramp_data_dir='.',
         the name of the submission to be tested
     """
     problem = assert_read_problem(ramp_kit_dir)
-    assert_title(ramp_kit_dir)
+
+    _print_title('Testing {}'.format(problem.problem_title))
+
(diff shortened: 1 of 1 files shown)


kegl:
The reason I did it this way is because when people start building a problem.py file, they may not have all the elements present yet. They may just want to call assert_cv without loading the data at the point where they call assert_cv from. In general, I tried to decouple these calls for a prize of a little overhead.

jorisvandenbossche:
> they may not have all the elements present yet. They may just want to call assert_cv without loading the data at the point where they call assert_cv from.

But the problem is that `assert_cv` actually loads the data as well, so I think the above will not be possible anyway?

kegl:
I'm suggesting that assert_cv optionally loads the data if it doesn't receive it.

PR #114, opened 2018-02-14
[MRG] Refactor tests

### LINKED ISSUE #71: Update fetch kit for conda integration
The kit fetching is installing the package via pip. in case that we want to use conda, we should probably think about it.

For instance: https://github.com/ramp-kits/california_rainfall/blob/master/environment.yml

### LINKED ISSUE #92: Travis fails with segfault
For the past 4 days or so the builds have been consistently failing with a segfault
```
Installing collected packages: tensorflow-tensorboard, tensorflow
  Found existing installation: tensorflow-tensorboard 0.4.0rc2
    Can't uninstall 'tensorflow-tensorboard'. No files were found to uninstall.
  Found existing installation: tensorflow 1.4.0
    Can't uninstall 'tensorflow'. No files were found to uninstall.
Successfully installed tensorflow-1.3.0 tensorflow-tensorboard-0.1.8
/home/travis/.travis/job_stages: line 57:  3267 Segmentation fault      (core dumped) pytest -s -v --cov=rampwf rampwf
The command "pytest -s -v --cov=rampwf rampwf" exited with 139.
cache.2
store build cache
```
Apparently it has to do with the `tensorflow` installation but I dont understand the segfault here.

Could it be some **conda vs. pip** issue ? 
What were the recent changes made to the kits or the libs for such issue to arise ?
@mehdidc @glemaitre @jorisvandenbossche @kegl

glemaitre:
closes #92
closes #71


jorisvandenbossche:
From the ones that are not yet included here:

'kaggle_seguro',
                        'MNIST_simplified',
                        'MNIST',
                        'california_rainfall_test',
                        'titanic',
                        'epidemium2_cancer_mortality',
                        'HEP_tracking',
                        'mouse_cytometry',

which ones are needed to include? Which ones have a unique workflow that is not yet covered by one of the included kits?

agramfort:
diff looks ok to me. Can you see the coverage report :

https://codecov.io/gh/paris-saclay-cds/ramp-workflow/pull/114?src=pr&el=continue

cf testing file in red

glemaitre:
We need to test the `predict` with the classifier (drug_spectra starting kit and iris starting kit).

glemaitre:
Do we need to have `drug_spectra` to be tested. Since that this is just a classifier and a regressor, if there is a kit using a classifier and another using regressor, including this kit will just be redundant.

@kegl Is there something specific in the kit to be tested?

jorisvandenbossche:
> We need to test the predict with the classifier (drug_spectra starting kit and iris starting kit).

Can you clarify this? They are both included.

> Do we need to have drug_spectra to be tested. Since that this is just a classifier and a regressor, if there is a kit using a classifier and another using regressor, including this kit will just be redundant.

The drug_spectra is a specific workflow. It's indeed consisting of a classifier and regressor, but it has a bit of custom code to glue them together, and this is something that still needs to be tested.

kegl:
> Do we need to have drug_spectra to be tested. Since that this is just a classifier and a regressor, if there is a kit using a classifier and another using regressor, including this kit will just be redundant.

> @kegl Is there something specific in the kit to be tested?

Yes, the drug_spectra workflow. As an alternative, we can put the workflow in the problem.py of drug_spectra (since it's quite specific).

kegl:
> From the ones that are not yet included here:

> 'kaggle_seguro',
> 'MNIST_simplified',
> 'MNIST',
> 'california_rainfall_test',
> 'titanic',
> 'epidemium2_cancer_mortality',
> 'HEP_tracking',
> 'mouse_cytometry',

> which ones are needed to include? Which ones have a unique workflow that is not yet covered by one of the included kits?

From memory, these ones only test scores, so their scores could just be added to iris and boston housing:

'kaggle_seguro',
'mouse_cytometry'

These ones test workflows and prediction types:

'MNIST_simplified',
'MNIST',
'california_rainfall_test',
'HEP_tracking',

'epidemium2_cancer_mortality' is a multi-column regression, so I think we should just add all regression scores to it which are supposed to work on multi-target regression.

kegl:
brier score caught a strange bug: probabilities are . 1.05; see current travis
it’s an sklearn random forest on iris
what’s the policy here?
should we normalize probabilities in every score?

kegl:
I added titanic because we need a binary classification problem for eg ROC AUC

kegl:
Is there a reason we don’t have a simple parameterized test_kit and a list of kits, as before?

kegl:
The pickling error still persists.

glemaitre:
> brier score caught a strange bug: probabilities are . 1.05; see current travis
it’s an sklearn random forest on iris
what’s the policy here?
should we normalize probabilities in every score?

Normally the score should be normalize. I will check that once that I finish a bit of cleaning with the nbconvert.

glemaitre:
> Is there a reason we don’t have a simple parameterized test_kit and a list of kits, as before?
I think that we should do it.

glemaitre:
I see something strange with the unicode. I thought that it has been fixed.
ping @aboucaud

kegl:
>Normally the score should be normalize. I will check that once that I finish a bit of cleaning with the nbconvert.

There was a conflict on my side in nbconvert, not sure why. Please check.

glemaitre:
I will work on a branch otherwise we will have trouble. But I will try to advance as much as possible this afternoon on those issues.

kegl:
I was doing coverage, thinking about doing blending for all prediction types (even though normally combine is also used in the cv-bagged score so it's a bit of over coverage), and adding the missing kits, also missing scores to iris, titanic, and boston. Is this ok, can we do this in parallel?

glemaitre:
> I was doing coverage, thinking about doing blending for all prediction types (even though normally combine is also used in the cv-bagged score so it's a bit of over coverage), and adding the missing kits, also missing scores to iris, titanic, and boston. Is this ok, can we do this in parallel?

yep I will play more with the maintenance for now. So it will not be an issue.

kegl:
Is it OK if I streamline the test a bit, writing a single script and a list of kits to test?

glemaitre:
> Is it OK if I streamline the test a bit, writing a single script and a list of kits to test?

We can always optimize it after ;)

glemaitre:
> brier score caught a strange bug: probabilities are . 1.05; see current travis
it’s an sklearn random forest on iris
what’s the policy here?
should we normalize probabilities in every score?

It is weird. It seems that it was a side effect from some install. I just remove the pinning of tensorflow and the error disappear.

glemaitre:
I added appveyor

rampwf/kits/ramp_kit.py
```diff
@@ -8,21 +8,22 @@
...
+#    #'boston_housing',
+#    #'iris',
+#    'titanic',
+#    'epidemium2_cancer_mortality',
+#    #'drug_spectra',
+#    #'air_passengers',
+#    #'el_nino',
+#    'HEP_tracking',
+#    'mouse_cytometry',
+#    )
```
agramfort: Cleanup?

rampwf/tests/kits/air_passengers/problem.py
```diff
@@ -0,0 +1,37 @@
+import os
+import pandas as pd
+import rampwf as rw
+from sklearn.model_selection import ShuffleSplit
```
agramfort: Wrong order

Rampwf should be imported last

rampwf/tests/kits/air_passengers/submissions/starting_kit/feature_extractor.py
```diff
@@ -0,0 +1,31 @@
+import pandas as pd
+import os
```
agramfort: Wrong import order

rampwf/tests/kits/boston_housing/problem.py
```diff
@@ -0,0 +1,39 @@
+import os
+import pandas as pd
+import rampwf as rw
+from sklearn.model_selection import ShuffleSplit
```
agramfort: Idem

rampwf/tests/kits/drug_spectra/problem.py
```diff
@@ -0,0 +1,83 @@
...
+"""
+
+# Author: Balazs Kegl <balazs.kegl@gmail.com>
+# License: BSD 3 clause
+
+import os
+import numpy as np
+import pandas as pd
+import rampwf as rw
+from sklearn.model_selection import ShuffleSplit
```
agramfort: Idem

rampwf/kits/ramp_kit.py
glemaitre: I think that we will just remove this module in fact

rampwf/tests/kits/titanic/data/test.csv
```diff
@@ -0,0 +1,90 @@
...
+655,0,3,"Hegarty, Miss. Hanora ""Nora""",female,18.0,0,0,365226,6.75,,Q
+392,1,3,"Jansson, Mr. Carl Olof",male,21.0,0,0,350034,7.7958,,S
+730,0,3,"Ilmakangas, Miss. Pieta Sofia",female,25.0,1,0,STON/O2. 3101271,7.925,,S
+856,1,3,"Aks, Mrs. Sam (Leah Rosen)",female,18.0,0,1,392091,9.35,,S
+32,1,1,"Spencer, Mrs. William Augustus (Marie Eugenie)",female,,1,0,PC 17569,146.5208,B78,C
+207,0,3,"Backstrom, Mr. Karl Alfred",male,32.0,1,0,3101278,15.85,,S
+671,1,2,"Brown, Mrs. Thomas William Solomon (Elizabeth Catherine Ford)",female,40.0,1,1,29750,39.0,,S
+384,1,1,"Holverson, Mrs. Alexander Oskar (Mary Aline Towner)",female,35.0,1,0,113789,52.0,,S
+496,0,3,"Yousseff, Mr. Gerious",male,,0,0,2627,14.4583,,C
+76,0,3,"Moen, Mr. Sigurd Hansen",male,25.0,0,0,348123,7.65,F G73,S
```
agramfort: shouldn't we compress the files in the repo? eg with gzip usring eg:

`df = pd.read_csv('data.csv.gz', compression='gzip')`

it would limit the repo size and allow to more easily push to pypi

glemaitre: Regarding PyPi, we don't include the test files I think right now.
It also mean that people installing the package from pypi will not be able to run the tests because some file will be missing.

I would think that it could be good to add them in the package even if we have to compress them.

agramfort: ok fair enough. Just try to make it as small as possible.

glemaitre: Only `el_nino` data are more than a 1MB (3MB to be precise). I would not bother I think :)

glemaitre:
> Is it OK if I streamline the test a bit, writing a single script and a list of kits to test?

@kegl I parametrize the tests. Like this it will automatically test any additional new kits and submissions.
I added a specific test for iris since that it seems you wanted to test the blending.

glemaitre:
So for me there is 2 issues to address:

* pickling does not work
* unicode when we pretty print

glemaitre:
@kegl @aboucaud @agramfort @jorisvandenbossche I would almost be in favor of merging it now. The changes are a lot already. So we could open 3 PRs for the remaining thing.

jorisvandenbossche:
I think we still need to add:

'MNIST_simplified',
'MNIST',
'california_rainfall_test',
'HEP_tracking',

?

kegl:
Yes, @jorisvandenbossche @glemaitre, those four are exactly what should still be added for getting the coverage back.

kegl:
Plus the Brier scores are still commented out in iris/problem.py because they fail. The random forest returns probabilities that are > 1 ( which is weird) and these scores don't normalize the probabilities. I did the transformation in https://github.com/paris-saclay-cds/ramp-workflow/blob/master/rampwf/score_types/soft_accuracy.py, lines 29, 31, and 40. Line 40 seemed the easiest solution to catch all 0 y_proba vectors, but it's specifict to this score. It should probably precede the score calculation, e.g. by setting all 0 y_proba to a uniform distribution over classes.

Maybe we should think about having these as a generic utility in https://github.com/paris-saclay-cds/ramp-workflow/blob/master/rampwf/score_types/classifier_base.py so these conversions are done before scores' `__call__` are called.

PR #116, opened 2018-02-16
Detection predictions blending based on clustering sampled gaussians

jorisvandenbossche:
This already adds the one based on HDBSCAN.

### DIFF
--- a/rampwf/prediction_types/detection.py
+++ b/rampwf/prediction_types/detection.py
@@ -34,7 +34,8 @@ class Predictions(BasePrediction):
         pass
 
     @classmethod
-    def combine(cls, predictions_list, index_list=None, greedy=False):
+    def combine(cls, predictions_list, index_list=None, greedy=False,
+                method=None):
 
         if index_list is None:  # we combine the full list
             index_list = range(len(predictions_list))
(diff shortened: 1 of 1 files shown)


rampwf/prediction_types/detection.py
```diff
@@ -243,20 +243,36 @@ def _sample_predictions(preds, n_samples=100):
...
+    labels = km.fit_predict(data)
+
+    objs = []
+
+    for i in range(n_clusters):
+        mask = labels == i
+        c = mask.sum() / max_n_samples
+        x, y = km.cluster_centers_[i]
+        r = np.mean((data[mask] - km.cluster_centers_[i])**2)
+        objs.append((c, x, y, r))
```
aboucaud: Why the choice of average of squares instead of the Euclidean distance ?

rampwf/prediction_types/detection.py
```diff
@@ -243,20 +243,36 @@ def _sample_predictions(preds, n_samples=100):
...
+    n_clusters = len(np.unique(labels[labels >= 0]))
 
     objs = []
 
     for i in range(n_clusters):
         mask = labels == i
         c = mask.sum() / max_n_samples
         x, y = data[mask].mean(axis=0)
-        r = np.sqrt(np.mean((data[mask] - np.array([x, y]))**2)) * 2
+        r = np.mean((data[mask] - np.array([x, y]))**2)
```
aboucaud: Same here

rampwf/prediction_types/detection.py
jorisvandenbossche: You mean for the calculation of the radius `r` ? Because I initially used the radius as the variance of multivariate normal that I sampled from. So therefore here using that as an estimate of the radius.

Issue #118, opened 2018-02-23
Follow-up on adding the blending code


jorisvandenbossche:
I merged https://github.com/paris-saclay-cds/ramp-workflow/pull/95, follow-up to do:

- add tests for blending
- restructure/refactor testing.py

PR #119, opened 2018-03-05
Refactoring of testing.py

aboucaud:
PR related to #118 and aiming at the clarification of
the `testing.py` for maintenance purposes.


rampwf/utils/prediction.py
```diff
@@ -0,0 +1,82 @@
+# coding: utf-8
```
kegl: Maybe prediction_io for the filename?

rampwf/utils/score.py
```diff
@@ -0,0 +1,112 @@
+# coding: utf-8
```
kegl: Maybe scoring.py rather than score.py?

rampwf/utils/colors.py
```diff
@@ -0,0 +1,80 @@
+# coding: utf-8
```
kegl: Maybe pretty_print.py?

PR #128, opened 2018-04-19
FIX use balanced accuracy from scikit-learn

### LINKED ISSUE #127: Bug in balanced accuracy
I am running the balanced accuracy in the autism kit and I get negative score.
I would think that we can future port the implementation from scikit-learn.

Do we have real test for the metrics which are not from scikit-learn?

glemaitre:
closes #127

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -22,7 +22,6 @@ install:
   # Add Library/bin directory to fix issue
   # https://github.com/conda/conda/issues/1753
   - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PYTHON%\\Library\\bin;%PATH%"
-  - conda update conda -y
   - conda create -n testenv --yes ipykernel pip python=%PYTHON_VERSION_ENV%
   - activate testenv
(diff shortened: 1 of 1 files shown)


glemaitre:
In scikit-learn, they consider this score meaningful only in binary classification. So it will not work for multiclass.

glemaitre:
In fact, this the reason it is failing on iris (more over iris is fully balanced so not sure it is meaninful to use it there)

glemaitre:
And the discussion to make it multiclass is there:
https://github.com/scikit-learn/scikit-learn/pull/10587

kegl:
I don't want to get into these debates now, just make sure it is backward compatible. It would probably be better to create a new score, like `balanced_accuracy_sk` or something like that.

glemaitre:
I checked. It is back compatible for multi-class apparently and does not give bad results for the binary case. But I would need a bit more time to check.

glemaitre:
Basically the only thing is to expose the `adjusted` parameter. `True` will fall back to the previous implementation while `False` will not adjust for chance.

PR #132, opened 2018-04-28
save fold scores

mehdidc:
Save the scores of each fold into a csv. I would find that useful when experimenting  with ramp_test_submission locally. Like I am thinking to have a minimal tool that displays the leaderboard based on all submissions in submission/.


kegl:
I like this, anybody has an objection?

agramfort:
not me

mehdidc:
Here is the tool if you want to take I look: https://github.com/mehdidc/ramp-leaderboard

kegl:
I like the additional tool, can you add it to rampwf and document it in the wiki?

PR #140, opened 2018-05-15
Allowing no test

kegl:


kegl:
We simply allow `get_test_data()` in `problem.py` to return `None, None` in case there is no test file. It could be useful even in the autism challenge if they want to reproduce the backend public leaderboard.

rampwf/tests/kits/titanic_no_test/titanic_no_test_starting_kit.ipynb
```diff
@@ -0,0 +1,1337 @@
+{
```
jorisvandenbossche: no need to add the notebook here I think

kegl: True, but doesn't the test need it?

aboucaud:
@kegl please fix flake8
```
rampwf/tests/test_kits.py:61:1: E303 too many blank lines (3)
```

PR #148, opened 2018-09-30
modification in RMSE to for periodic target

yetkinyilmaz:
RMS not just by the difference but rolls for 2*pi (or other limits) when necessary…


yetkinyilmaz:
please let me know if you'd like me to implement in any other way, and if I should also do it for other error types...

yetkinyilmaz:
canceling this, to be re-submitted from up-to-date base

PR #149, opened 2018-09-30
modification in RMSE to for periodic target

yetkinyilmaz:
RMS not just by the difference but rolls for 2*pi when necessary…


kegl:
Please put the periodic RMSE in a new score. You can inherit from RMSE if you want.

The travis error is due to the new sklearn.

yetkinyilmaz:
correction: not derived, just copied... it's an independent score now.

yetkinyilmaz:
Canceled: Moved into problem.py in ramp-kits/mechanics until more general need...

PR #155, opened 2018-10-04
Handle DataFrame y_pred/y_true in run_submission_on_cv_fold

jorisvandenbossche:
Not sure if this should be merged, but this is needed to run https://github.com/ramp-kits/solar_wind/pull/4


kegl:
This would require a major rewrite of the server backend on databoard. It's not impossible but we can't do it within the timeframe.

jorisvandenbossche:
OK, no problem, this should be solvable on the solar wind side (either we just drop the time information, and don't take it into account for the scoring, or either I store the datetimes as integers in an array, and reconstruct the dataframe for scoring)

PR #158, opened 2018-10-10
lazy import tabulate

kegl:
The fancy new local `ramp_leaderboard` uses tabulate that doesn't work on windows. We'll catch the exception and revert to pandas rendering the table.


kegl:
@jorisvandenbossche could you please check appveryor. It fails on python 3.*.

jorisvandenbossche:
Running the titanic notebook is timing out apparently

jorisvandenbossche:
Will open a separate issue about that

PR #161, opened 2018-10-19
saving state and error, not only y_pred

kegl:
To unify local and remote training, we'll save everything that we need in the DB into files in `submissions/<submission>/training_output/fold_<i>`, including the state. After rsync, the script can check the file `state.txt` (in each fold), and act accordingly. If the state is `training_error`, `validation_error`, or `test_error`, `submissions/<submission>/training_output/fold_<i>/error.txt` will contain part of the stack that corresponds to the user submission.

### DIFF
--- a/rampwf/utils/command_line.py
+++ b/rampwf/utils/command_line.py
@@ -144,6 +144,11 @@ def create_ramp_blend_submissions_parser():
                         type=str,
                         help='Directory containing the data. This directory'
                         ' should contain a "data" folder.')
+    parser.add_argument('--ramp_submission_dir',
+                        default='submissions',
+                        type=str,
+                        help='Directory where the submissions are stored. It '
+                        'should contain a "submissions" directory.')
     parser.add_argument('--submissions',
                         default='ALL',
                         type=str,
@@ -181,6 +186,7 @@ def ramp_blend_submissions():
 
     blend_submissions(ramp_kit_dir=args.ramp_kit_dir,
                       ramp_data_dir=args.ramp_data_dir,
+                      ramp_submission_dir=args.ramp_submission_dir,
                       submissions=submissions,
                       save_output=save_output,
                       min_improvement=float(args.min_improvement))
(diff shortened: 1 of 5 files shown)


rampwf/utils/command_line.py
```diff
@@ -45,10 +50,10 @@ def create_ramp_test_submission_parser():
     parser.add_argument('--pickle', dest='pickle', action='store_true',
                         help='Specify this flag to pickle the submission '
                              'after training.')
-    parser.add_argument('--save-y-preds', dest='save_y_preds',
+    parser.add_argument('--save-output', dest='save_output',
```
aboucaud: Since you don't want to break things, maybe you should keep `--save-y-preds` for now but make it point to 
the `save_output` arg, i.e.
```python
parser.add_argument('--save-y-preds', dest='save_output',
...
```
We will remove this argument once we've made sure `--save-y-preds` has be removed from all of the starting kits/docs/wikis.

mehdidc:
It looks good, I would just say that we would need to set  the state as 'training'  just before calling train_submission in case we have a loop, in order to  distinguish between non trained submissions and submissions being trained, no ?

rampwf/utils/command_line.py
kegl: OK, try it.

kegl:
Ok @mehdidc, can you add this?

kegl:
Ok, I did the change @mehdi proposed. Note that state is now defined for each fold, not for the full submission. We have this distinction also in the db, but I'm wondering how should we do this here. Either we derive submission state by reading the fold states in a script, or we explicitly set the submission state.

rampwf/utils/command_line.py
kegl: OK, done, it seems to work.

rampwf/utils/command_line.py
```diff
@@ -50,10 +50,15 @@ def create_ramp_test_submission_parser():
...
```
glemaitre: For consistency with the other parameter, it should be `--save_ouptut` actually.

However, I think that we should refactor using `click` and all parameters will used `-` instead of `_`.
It might be worth to open an issue for `click` refactoring.

This is actually breaking the backend worker as well.

PR #166, opened 2018-12-17
[MRG] EHN: add submission path when to train a submission

glemaitre:
This PR allows to pass a new parameter `ramp_submission_dir` to specify a a directory where the submissions are stored. This is useful when dealing with the backend server.


glemaitre:
@kegl could you have a look at this PR.
We have a failure in windows but this is a timeout on the conversion of notebook.

PR #171, opened 2019-01-09
Score reorder bug

kegl:
Score matrix was reordered when doing `unstack()` so precisions specified in `problem.score_types` were applied in the wrong order before printing. I added an explicit `reorder_df_scores` which is called before rounding and printing. In addition, I did some naming cosmetics in `bag_submissions` (ie `score_types` instead of `score_type` and `score_type` instead of `score`) to be consistent.

### DIFF
--- a/rampwf/utils/scoring.py
+++ b/rampwf/utils/scoring.py
@@ -4,10 +4,36 @@ Scoring utilities
 """
 import numpy as np
 import pandas as pd
+from .pretty_print import print_warning
+
+
+def reorder_df_scores(df_scores, score_types):
+    """Reorder scores according to the order in score_types.
+
+    Parameters
+    ----------
+    df_scores : pd.DataFrame
+        the score dataframe
+    score_types : list of score types
+
+    Returns
+    -------
+    df_scores : the dataframe with reordered scores
+    """
+    try:
+        # try to re-order columns/rows in the printed array
+        # we may not have all train, valid, test, so need to select
+        index_order = np.array(['train', 'valid', 'test'])
+        ordered_index = index_order[np.isin(index_order, df_scores.index)]
+        df_scores = df_scores.loc[
+            ordered_index, [score_type.name for score_type in score_types]]
+    except Exception:
+        print_warning("Couldn't re-order the score matrix..")
+    return df_scores
 
 
 def mean_score_matrix(df_scores_list, score_types):
-    """Construct a mean ± std score dataframe from a list of score dataframes.
+    u"""Construct a mean ± std score dataframe from a list of score dataframes.
 
     Parameters
     ----------
@@ -58,6 +84,7 @@ def score_matrix_from_scores(score_types, steps, scoress):
     df_scores = pd.DataFrame(results)
     df_scores = df_scores.set_index(['step', 'score'])['value']
     df_scores = df_scores.unstack()
+    df_scores = reorder_df_scores(df_scores, score_types)
     return df_scores
 
 
--- a/rampwf/utils/submission.py
+++ b/rampwf/utils/submission.py
@@ -338,11 +338,10 @@ def bag_submissions(problem, cv, y_train, y_test, predictions_valid_list,
         test_idx = ([valid_is for (train_is, valid_is) in cv]
                     if step == 'valid' else None)
         score_dict = {}
-        for score_type in score_types:
+        for st in score_types:
             pred, scores = get_score_cv_bags(
-                score_type, pred_list, gt_list, test_is_list=test_idx
-            )
-            score_dict[score_type.name] = {
+                st, pred_list, gt_list, test_is_list=test_idx)
+            score_dict[st.name] = {
                 key: val for key, val in enumerate(scores)}
         bagged_scores[step] = score_dict
         # the predictions will always be the same for all score and we store
--- a/rampwf/utils/submission.py
+++ b/rampwf/utils/submission.py
@@ -7,14 +7,13 @@ import os
 from collections import Iterable
 from collections import OrderedDict
 
-import numpy as np
 import pandas as pd
 import cloudpickle as pickle
 
 from .io import save_y_pred
 from .combine import get_score_cv_bags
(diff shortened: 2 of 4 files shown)


rampwf/utils/submission.py
```diff
@@ -320,9 +320,10 @@ def bag_submissions(problem, cv, y_train, y_test, predictions_valid_list,
     print_title('----------------------------')
     score_type_index = (slice(None) if score_type_index is None
                         else score_type_index)
-    score_type = problem.score_types[score_type_index]
-    score_type = ([score_type] if not isinstance(score_type, Iterable)
-                  else score_type)
+    score_types = problem.score_types[score_type_index]
+    score_types = (
+        [score_types] if not isinstance(score_types, Iterable)
+        else score_types)
```
glemaitre: You should either let it as it is or use

```python
    score_types = (
        [score_types]
        if not isinstance(score_types, Iterable)
        else score_types
)
```
if you want to use black or yapf style

rampwf/utils/submission.py
```diff
@@ -337,12 +338,12 @@ def bag_submissions(problem, cv, y_train, y_test, predictions_valid_list,
         test_idx = ([valid_is for (train_is, valid_is) in cv]
                     if step == 'valid' else None)
         score_dict = {}
-        for score in score_type:
+        for score_type in score_types:
```
glemaitre: I am against this change: the distinction between the plural and singular is so close that it is prone to error and to exchange them in the code. I am not against using the plural thought:


```python
for st in score_types:
```
would be less prone to error

rampwf/utils/submission.py
```diff
@@ -337,12 +338,12 @@ def bag_submissions(problem, cv, y_train, y_test, predictions_valid_list,
...
             pred, scores = get_score_cv_bags(
-                score, pred_list, gt_list, test_is_list=test_idx
+                score_type, pred_list, gt_list, test_is_list=test_idx
             )
-            score_dict[score.name] = {key: val
-                                      for key, val in enumerate(scores)}
+            score_dict[score_type.name] = {
+                key: val for key, val in enumerate(scores)}
```
glemaitre: The yapf or black style will be then

```python
score_dict[score_type.name] = {
                key: val
                for key, val in enumerate(scores)
}
```

In this case we can specify what the key in val instead of the general naming.

rampwf/utils/submission.py
```diff
@@ -355,22 +356,21 @@ def bag_submissions(problem, cv, y_train, y_test, predictions_valid_list,
...
     # prepare the bagged scores which will be printed.
-    numerical_precision = pd.Series([score.precision for score in score_type],
-                                    index=[score.name for score in score_type])
     highest_level = df_scores.index.get_level_values('n_bag').max()
-    df_scores_rounded = df_scores.loc[(slice(None), highest_level), :].round(
-        numerical_precision
-    )
-    df_scores_rounded.index = df_scores_rounded.index.droplevel('n_bag')
-    print_df_scores(df_scores_rounded, score_type, indent='\t')
+    df_scores = df_scores.loc[(slice(None), highest_level), :]
```
glemaitre: You should not do that

df_scores corresponds to the dataframe which will be dumped on the disk. Therefore, it contains all bags and all scores information.

df_scores_rounded takes only the last bag, round it and print it.

Since the saving is coming after the printing, we do not store all the information which are expected by the database.

rampwf/utils/submission.py
```diff
@@ -355,22 +356,21 @@ def bag_submissions(problem, cv, y_train, y_test, predictions_valid_list,
...
+    df_scores.index = df_scores.index.droplevel('n_bag')
+    df_scores = reorder_df_scores(df_scores, score_types)
+    df_scores = round_df_scores(df_scores, score_types)
+    print_df_scores(df_scores, indent='\t')
 
     if save_y_preds:
-        bagged_scores_filename = os.path.join(training_output_path,
-                                              'bagged_scores.csv')
+        bagged_scores_filename = os.path.join(
+            training_output_path, 'bagged_scores.csv')
```
glemaitre: same deal about the styling.

rampwf/utils/scoring.py
```diff
@@ -4,10 +4,36 @@
...
+    score_types : list of score types
+
+    Returns
+    -------
+    df_scores : the dataframe with reordered scores
+    """
+    try:
+        # try to re-order columns/rows in the printed array
+        # we may not have all train, valid, test, so need to select
+        index_order = np.array(['train', 'valid', 'test'])
```
glemaitre: no need for numpy here

```python
index_order = ['train', 'valid', 'test']
index_order = sorted(set(index_order) & set(df_scores.index),
                                  key=index_order.index)

```

rampwf/utils/scoring.py
```diff
@@ -4,10 +4,36 @@
...
+        df_scores = df_scores.loc[
+            ordered_index, [score_type.name for score_type in score_types]]
+    except Exception:
+        print_warning("Couldn't re-order the score matrix..")
+    return df_scores
 
 
 def mean_score_matrix(df_scores_list, score_types):
-    """Construct a mean ± std score dataframe from a list of score dataframes.
+    u"""Construct a mean ± std score dataframe from a list of score dataframes.
```
glemaitre: do not use `"u"` you should use `r"""` and the latex notation ``:math:`\pm` ``

rampwf/utils/submission.py
kegl: OK.

rampwf/utils/submission.py
kegl: OK.

rampwf/utils/submission.py
kegl: OK.

rampwf/utils/scoring.py
kegl: OK

rampwf/utils/scoring.py
kegl: OK.

PR #175, opened 2019-01-20
MAINT: refactor the CLI using click

### LINKED ISSUE #173: Use click for the CLI tools
For consistency with the other CLI in the RAMP bundle, we should use `click`.

glemaitre:
closes #173 

TODO:

- [ ] CLI for tests
- [ ] CLI for covnert
- [ ] CLI for show
- [ ] Write unit tests

### DIFF
--- a/rampwf/utils/cli/testing.py
+++ b/rampwf/utils/cli/testing.py
@@ -8,25 +8,22 @@ from ..testing import assert_submission
 CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
 
 
-@click.group(context_settings=CONTEXT_SETTINGS)
-def main():
-    """Command-line to test submission and notebook before to submit."""
-    pass
-
-
-@main.command()
-@click.option('--ramp-kit-dir', default='.',
+@click.command(context_settings=CONTEXT_SETTINGS)
+@click.option('--ramp-kit-dir', default='.', show_default=True,
               help='Root directory of the ramp-kit to test.')
-@click.option('--ramp-data-dir',  default='.',
+@click.option('--ramp-data-dir', default='.', show_default=True,
               help='Directory containing the data. This directory should '
               'contain a "data" folder.')
 @click.option('--ramp-submission-dir', default='submissions',
+              show_default=True,
               help='Directory where the submissions are stored. It should '
               'contain a "submissions" directory.')
-@click.option('--submission', default='starting_kit',
+@click.option('--submission', default='starting_kit', show_default=True,
               help='The kit to test. It should be located in the '
               '"submissions" folder of the starting kit. If "ALL", all '
               'submissions in the directory will be tested.')
+@click.option('--notebook', is_flag=True, show_default=True,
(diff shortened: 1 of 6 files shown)


rampwf/utils/cli/testing.py
```diff
@@ -0,0 +1,81 @@
...
+@click.option('--pickle', is_flag=True,
+              help='Specify this flag to pickle the submission after '
+              'training.')
+@click.option('--save-output', is_flag=True,
+              help='Specify this flag to save predictions, scores, eventual '
+              'error trace, and state after training.')
+@click.option('--retrain', is_flag=True,
+              help='Specify this flag to retrain the submission on the full '
+              'training set after the CV loop.')
+def submission(ramp_kit_dir, ramp_data_dir, ramp_submission_dir, submission,
```
jorisvandenbossche: I would not make this a subcommand of main, but just make this a one-level command?
Because now you would need to do `ramp-test submission --submission name`. We could make it `ramp-test --submission name`, which is much less duplicative.

That would move the testing of the notebook to another command, but I think that is actually a good thing, as no end-user of `rampwf` should need this? (only ramp-kit authors)

rampwf/utils/cli/testing.py
```diff
@@ -0,0 +1,81 @@
...
+@main.command()
+@click.option('--ramp-kit-dir', default='.',
+              help='Root directory of the ramp-kit to test.')
+def notebook(ramp_kit_dir):
+    """Test a notebook  before to submit on RAMP studio."""
+    assert_notebook(ramp_kit_dir=ramp_kit_dir)
+
+
+def start():
+    main()
```
jorisvandenbossche: why is this actually needed? It is not possible to directly call `main`?

rampwf/utils/cli/testing.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@click.group(context_settings=CONTEXT_SETTINGS)
+def main():
+    """Command-line to test submission and notebook before to submit."""
+    pass
+
+
+@main.command()
+@click.option('--ramp-kit-dir', default='.',
```
jorisvandenbossche: The defaults are not shown in the help, which is a bit annoying. Do you know of any way how to do this with click?

glemaitre: Apparently there is a show_defaults=True for that

rampwf/utils/cli/testing.py
glemaitre: I don't have objection. We could actually make both testing in the same function and perform whatever testing if you specify notebook, submission or both

kegl: I agree. But I'd rather do `ramp sdk test_submission --submission`. `test` is a bit too generic for me and it suggests a code test, not an ML model test.

jorisvandenbossche:
Appveyor was already passing before, so merging.

There are still some TODOs: add tests + deprecate current testing commands? + convert remaining commands to click

PR #177, opened 2019-02-13
[WIP] hyperopt

kegl:


rampwf/hyperopt/tests/interfaces/header_in_files/titanic/submissions/starting_kit/classifier.py
```diff
@@ -4,24 +4,19 @@
 from sklearn.base import BaseEstimator
 
 # RAMP START HYPERPARAMETERS
-impute_strategy = 'mean'  # opt: ['median', 'mean']
-logistic_C = 0.01  # opt: [0.001, 0.01, 0.1, 0.5, 0.9, 0.99, 1.0]
-# RAMP END HYPERPARAMETERS
-
+hyper_parameters = {'logreg_C' : 1,
+                    'imputer_strategy' : 'median'}
+# RAMP END
```
kegl: I would keep `RAMP END HYPERPARAMETERS` for better human readbility.

rampwf/utils/testing.py
```diff
@@ -234,3 +242,91 @@ def blend_submissions(submissions, ramp_kit_dir='.', ramp_data_dir='.',
...
+        submission_name = os.path.split(submission_path.rstrip(os.path.sep))[1]
+        temp_submission_path = os.path.join(dirname, submission_name)
+        shutil.copytree(submission_path, temp_submission_path)
+
+        _configure_submission(params, problem, temp_submission_path, dirname)
+
+        df_scores_list = []
+        predictions_valid_list = []
+
+        for fold_i, fold in enumerate(problem.get_cv(X_train, y_train)):
```
kegl: The hyperopt engine should have a granularity of running `params` on each cv fold separately, ie, being able to decide how many folds it needs. On a very bad combination we may need to run only a single fold, to save time. On the single fold results we should then store: `fold_i`, `n_train`, `n_test`, `score`, and even `train_time` and `test_time` (or even `RAM` if we have it). So the hyperopt algo has all the information to smartly choose the next params and `fold_i`.

rampwf/hyperopt/tests/interfaces/header_in_files/titanic/submissions/starting_kit/classifier.py
```diff
@@ -0,0 +1,27 @@
...
+from sklearn.impute import SimpleImputer
+from sklearn.pipeline import Pipeline
+from sklearn.base import BaseEstimator
+from rampwf.hyperopt import Hyperparameter
+
+# RAMP START HYPERPARAMETERS
+logreg_C = Hyperparameter(
+    dtype='float', default=1.0, values=[0.01, 0.1, 0.9, 1.0])
+imputer_strategy = Hyperparameter(
+    dtype='object', default='median', values=['mean', 'median'])
```
agramfort: I feel these changes are too invasive for ramp-workflow and it tends to reinvent the notion of parameter space present in hyperopt and scikit-optimize (https://scikit-optimize.github.io/notebooks/hyperparameter-optimization.html). I would do something like this:

```
logreg_C = 1.0  # values : [0.01, 0.1, 0.9, 1.0]
imputer_strategy = 'median'  # values : ['mean', 'median']
```

by doing this hyperopt does not need yet to be in core ramp-workflow.
A first PR here would be to be able to parse the header, eventually
print a warning when not found.

to embrace more the hyperopt or scikit-optimize approach it could be:

```
logreg_C = 1.0  # Real(10**-5, 10**0, "log-uniform")
imputer_strategy = 'median'  # Categorical(['mean', 'median'])
```

kegl: We discussed the syntax options with Amine and decided to go with the pure python declaration for two reasons: 1) the python interpreter checks syntax instead of our parser and 2) it was easier to get the hypers by loading the file and filtering the Hyperparameter objects. But I don't have a strong opinion, if someone feels strongly enough to implement the comment option, we could add both and see which one is used.

On the  scikit-optimize/hyperopt way of defining spaces I have a stronger opinion, it's an overly complicated design. No reason to have continuous spaces, in my experience it's the data scientist's prior knowledge what values to try, but tell me the use case in which you really need continuous (ie you have no idea about the granularity/smoothness of the score wrt your hyper) and I change my mind. In this way you don't need distribution parameter either, just specify a logarithmic grid if you want. We also have an optional `prior` in the declaration that you can communicate to your hyperopt engine (random uses it already). Finally, another advantage is that we can use bandits, and even GP is much happier on a predefined grid (e.g. you can measure nuggets/noise easily).

agramfort:
>
> We discussed the syntax options with Amine and decided to go with the pure
> python declaration for two reasons: 1) the python interpreter checks syntax
> instead of our parser and 2) it was easier to get the hypers by loading the
> file and filtering the Hyperparameter objects. But I don't have a strong
> opinion, if someone feels strongly enough to implement the comment option,
> we could add both and see which one is used.
>
what I don't like is that we make another layer of complexity for RAMP
users. The closer we are to standards such as sklearn pipeline, a standard
def of parameter space the better. It means less things to document for
users to get started.

> On the scikit-optimize/hyperopt way of defining spaces I have a stronger
> opinion, it's an overly complicated design. No reason to have continuous
> spaces, in my experience it's the data scientist's prior knowledge what
> values to try, but tell me the use case in which you really need continuous
> (ie you have no idea about the granularity/smoothness of the score wrt your
> hyper) and I change my mind.
>
we can start with discrete values and do plain grid searches or random
search on fixed grids. But for a gradient boosting typically I would be
it's better to have intervals eg for learning rate.

> In this way you don't need distribution parameter either, just specify a
> logarithmic grid if you want. We also have an optional prior in the
> declaration that you can communicate to your hyperopt engine (random uses
> it already). Finally, another advantage is that we can use bandits, and
> even GP is much happier on a predefined grid (e.g. you can measure
> nuggets/noise easily).
>

hyperopt and skopt are not super happy with discrete values that have a
natural order I think. But maybe it's an implementation issue.

rampwf/hyperopt/hyperopt.py
```diff
@@ -0,0 +1,521 @@
...
+        self.df_summary_ = pd.merge(
+            counts, self.df_summary_, left_index=True, right_index=True)
+        print(self.df_summary_)
+        summary_fname = os.path.join(hyperopt_output_path, 'summary.csv')
+        self.df_summary_.to_csv(summary_fname)
+
+    def _save_best_model(self):
+        official_scores = self.df_summary_[
+            'valid_' + self.problem.score_types[0].name + '_m']
+        best_defaults = official_scores.idxmax()
```
albertcthomas: depending on `is_lower_the_better` value, this can be `idxmax()` or `idxmin()`

kegl: done, nice catch!

glemaitre:
CIs are broken

agramfort:
I would suggest to revert and reopen a PR with green Cis. Unless @kegl you can fix it quickly.

agramfort:
this PR broke the CIs 3 months ago @kegl 

2 options:
- revert this commit
- fix the CIs

deadline is in 8 days for me as the datacamp starts again and I want to avoid having to deal with students having messy / heterogeneous setups.

cc @lucyleeow

kegl:
Last time I checked the breaking of travis was independent of the commit. Reverting would not work.

PR #178, opened 2019-02-14
Add ability to specify worker in ramp-test submission

jorisvandenbossche:
This works with the master branch of ramp-board (only ramp-engine and ramp-utils is needed). 

The usage is like:

```
ramp-test --submission starting_kit_local2 --worker conda_config.yml 
# or
ramp-test --submission starting_kit_local2 --worker aws_config.yml 
```

where the configuration file then has the information about the `worker`.

TODO:

- [ ] add tests
- [ ] think about how this needs to interact with the other options to `ramp-test`, because currently they are not passed through.


jorisvandenbossche:
Need to update this for the latest changes in ramp-board

kegl:
Looks great! Yes, can't we just pass the other options to the `ramp-test` command that the worker executes?

PR #185, opened 2019-10-25
[MRG] [DOC] Add sphinx documentation

lucyleeow:
Update documentation and move from wiki to sphinx documentation.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,6 +3,9 @@ jobs:
   build:
     docker:
       - image: circleci/python:3.6-stretch
+    branches:
+      only:
+        - DOC
     steps:
       - checkout
       - run: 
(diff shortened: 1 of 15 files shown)


agramfort:
@lucyleeow where is this doc coming from? did you type all this? Can you see why CIs are red?

lucyleeow:
@agramfort yes I wrote this. @maikia and I thought it would be a good idea to write some (better) documentation to make it easier for us to write starting kits.

appveyor seems to be failing while running ``python setup.py egg_info`` - no idea why this is.
Travis is failing 1 test (for the el nino kit) for Python 2.7 (possibly due to deprecation?)

agramfort:
nice !

I would first do a PR to fix the CIs before merging this.

>

lucyleeow:
@agramfort to add circleCI, it would be easiest to merge this so master has a `.circleci/`.

appveyor failing is independent and we found it to be due to time out error when executing the titanic starting kit ipynb due to the `! ramp_test_submission` cell.

Merging.

PR #186, opened 2019-10-28
[wip] fixing issues with tests not passing on appveyor

maikia:


maikia:
Yes, the tests pass through. However, one of the lines which caused the problem (only on Windows, Python 3) was this one:
!ramp_test_submission 
(in the titanic and titanic_no_test jupyter notebooks). 
For now it is simply commented out allowing the tests to pass, however the problem is not solved (ie I don't know why this command is not being executed as it should). Do you have an idea @agramfort , @glemaitre , @lucyleeow @kchawla-pi @lesteve ?

kchawla-pi:
As we discussed IRL, it could be the shell command is not working the same on a windows shell. I will check this out at home on my windows system and report back here.

lucyleeow:
For reference the test that is failing is `test_notebook_testing` for the titanic starting kit notebook. Error is TimeoutError.

Specifically the fails at `! ramp_test_submission` cell. Works when this cell was removed.

Link: https://ci.appveyor.com/project/glemaitre/ramp-workflow/builds/28433296/job/t863yuqxmjdtlnqe

agramfort:
Lets merge then.

maikia:
Ok, I merged. Although still I would like to know why this is causing a problem

PR #194, opened 2019-12-02
Fixing blending

kegl:


kegl:
Anybody would like to review it?

This is just to update blending so we can use it in ramp-board (instead of the old code).

Travis passes except for some flake stuff in the detection score type tests.

Issue #196, opened 2019-12-05
Address warnings in 'titanic' and 'iris' kits


lucyleeow:
Fix deprecation warnings in starting_kit of titanic (simple fixes):
```
/home/lucy/miniconda3/envs/FAN/lib/python3.7/site-packages/sklearn/utils/deprecation.py:66: DeprecationWarning: Class Imputer is deprecated; Imputer was deprecated in version 0.20 and will be removed in 0.22. Import impute.SimpleImputer from sklearn instead.
  warnings.warn(msg, category=DeprecationWarning)
/home/lucy/miniconda3/envs/FAN/lib/python3.7/site-packages/sklearn/linear_model/logistic.py:432: FutureWarning: Default solver will be changed to 'lbfgs' in 0.22. Specify a solver to silence this warning.
  FutureWarning)
```

and this warning with the starting_kit of iris (not sure about this one):

```
/home/lucy/miniconda3/envs/FAN/lib/python3.7/site-packages/sklearn/metrics/classification.py:1437: UndefinedMetricWarning: F-score is ill-defined and being set to 0.0 in labels with no predicted samples.
  'precision', 'predicted', average, warn_for)
```
The changes should be copied over to 'ramp-kits' as well.

Happy to tackle this later unless someone else is interested.

PR #197, opened 2019-12-05
[DOC] Add doc images

lucyleeow:
I thought the images for the documentation was added in #195 but it's hard to tell now as that PR is a bit of a mess.


lucyleeow:
I think the failings are due to #196. Merging as this just adds some images.

PR #199, opened 2019-12-06
Leaderboard enh

kegl:
Adding bagged leaderboard to `ramp show leaderboard` as an option and making it default.


agramfort:
Can you add a test that asserts that what is obtained is ok?

kegl:
Hm, this is just a printing script. Or you mean fixing titanic? That seems to be caused by travis switching to sklearn 0.22 and it's part of the test already.

agramfort:
you add lines of code that are not covered by any test and for which we
don't assess if output is correct.
I am just saying that we should try to avoid this. Me as a reviewer I would
have to fetch, try etc to see if
it looks good before merging. Using CIs is more transparent.

>

rampwf/utils/cli/tests/test_cli.py
```diff
@@ -0,0 +1,81 @@
+# -*- coding: utf-8 -*-
```
agramfort: @kegl why not building on existing tests here https://github.com/paris-saclay-cds/ramp-workflow/blob/master/rampwf/utils/tests/test_command_line.py#L435 ?

kegl: Those test the old cli (`ramp_test_submission`). The new added by @glemaitre seemed easier for me to start with. It was untested, now it's covered. These are printing functions, no output, no other modules depend on this. I don't want to test either the content of the tables or the actual format. The current test protects against inadvertent modification of the printing routing.

We should perhaps deprecate the old cli. Two remarks:
1. The dispatcher still uses the old cli.
2. the command `ramp` is created only when ramp-board is installed, maybe ramp-workflow should also do it.

Issue #205, opened 2020-01-17
Master failing at digits classification


lucyleeow:
Failing on Travis for Python 3.6 at the test kit 'digits'.

Likely due to dependency version as the latest PR was green when merged.

cc @kegl @agramfort

agramfort:
I restarted the job. Let's see what happens

>

glemaitre:
I just change the title. I think that we need to migrate the notebook and project which are using `keras` to use `tensorflow.keras` and check what is it failing with tensorflow 2.1

kegl:
Is this still an issue? We have tensorflow 2.3 now.

nicolasbdls:
Hello,
I would like to submit a project on ramp using `keras`.
However when I submit my file, I get the error "No module names 'keras'" or "No module names 'tensorflow'" if I import `tensorflow.keras`.
Is there a solution to my problem? 
Thanks

PR #206, opened 2020-01-17
FIX failing test

lucyleeow:
#205

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ before_install:
 install:
     - conda create -n testenv --yes pip python=$PYTHON_VERSION ipykernel
     - source activate testenv
+    - pip install --upgrade pip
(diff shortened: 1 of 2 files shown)


lucyleeow:
Looking at the tensorflow system requirements, it seems a newer version of pip is require than what we are using:

```
$ pip --version

pip 9.0.1 from /home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages (python 3.6)
```

I'll update the `.travis.yml` file to to update pip, unless there are objections?

lucyleeow:
@agramfort will the underlying problem for the fail on tensorflow > 2.0.0 need to be addressed though?

agramfort:
it's a packaging tensorflow issue I would say.

>

lucyleeow:
I'll merge then.

PR #210, opened 2020-01-21
MNT remove deprecation warning due to our code

### LINKED ISSUE #204: Avoid deprecation warning
As the `imp` package is deprecated I suggest to now use `importlib` and change the [`import_file` function](https://github.com/paris-saclay-cds/ramp-workflow/blob/master/rampwf/utils/importing.py#L10) according to the [python doc](https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly).

Another possibility is to use the `runpy` module but this would imply to change the way the elements of a module are accessed.

glemaitre:
closes #204 

Remove `DeprecationWarning` due to imp and `collections.abc`
Updating `tabulate` in this regard.

### DIFF
--- a/rampwf/externals/tabulate.py
+++ b/rampwf/externals/tabulate.py
@@ -4,29 +4,37 @@
 
 from __future__ import print_function
 from __future__ import unicode_literals
-from collections import namedtuple, Iterable
+from collections import namedtuple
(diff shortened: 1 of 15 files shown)


albertcthomas:
Thanks @glemaitre

PR #213, opened 2020-01-23
BUG fix trace saving

### LINKED ISSUE #207: Exception trace index in `print_submission_exception`
Does anyone know what [this](https://github.com/paris-saclay-cds/ramp-workflow/blob/4bf2b0f0be205b9ff628e371961ab4df0d3e8ff5/rampwf/utils/io.py#L124) line in `print_submission_exception` is for?

```
    if trace[4].find('load_source') > -1:
        trace = trace[5:]
```
When I got an exception, there was a second exception which was that `trace[4]` was out of index (there were only 4 elements in `trace` not >=5). Why do we assume `trace` will be at least 5 elements?

lucyleeow:
closes #207

### DIFF
--- a/rampwf/utils/io.py
+++ b/rampwf/utils/io.py
@@ -125,7 +125,7 @@ def print_submission_exception(save_output, output_path):
     try:
         if trace[4].find('load_source') > -1:
             trace = trace[5:]
-    except:
+    except IndexError:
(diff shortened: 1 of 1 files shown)


lucyleeow:
cc @agramfort @glemaitre

PR #225, opened 2020-03-02
colored output only form terminals that know how to handle it

kegl:


albertcthomas:
I will test it on both cmder and anaconda prompt and let you know.

rampwf/utils/pretty_print.py
```diff
@@ -2,13 +2,14 @@
...
 """
-import platform
+import os
 
 from pandas import option_context
 
 from ..externals.colored import stylize, fg, attr
 
-IS_WINDOWS = platform.system() == "Windows"
+COLOR_TERMS = ['xterm-256color', 'cygwin', 'xterm-color']
```
agramfort:
```suggestion
COLOR_TERMS = ['xterm-256color', 'cygwin', 'xterm-color', 'xterm']
```

to make it work on my macos

kegl: Ah, that's a problem. I added 'xterm' first but it's also 'xterm' in the anaconda terminal on windows which has no handling of the escape sequences. On my mac (both in Terminal and in iterm) it's 'xterm-256color'. What should we do? Be more conservative or liberal?

kegl: I could make a more complex condition:
```
'TERM' in os.environ and (os.environ['TERM'] in COLOR_TERMS or (os.environ['TERM'] == 'xterm' and not IS_WINDOWS)
```

agramfort:
ok let's try this. We can adjust when we see people hitting the issue.
If it fixes issue for you, me, guillaume and windows users using anaconda prompt
it's already great !

kegl:
OK, @agramfort pls merge if you're happy. The AppVeyor failure seems unrelated.

agramfort:
works on my mac.

can someone else test? @albertcthomas does it work for you on windows wtih cmdr and anaconda prompt?

albertcthomas:
With last commit, same as before:

> This is working for me in cmder (with colors) and in command prompt/anaconda prompt/power shell (without colors).

agramfort:
great

let's wait for @glemaitre to test before we merge.

>

glemaitre:
Working in Linux. The failure is due to a missing wheel of scikit-learn for Python 3.5 (then we try to install from source which is not intended. We should probably update the support of ramp-workflow for 3.6/3.7 (not 3.8 because of tensorflow)

agramfort:
ok green. Thanks @kegl for the fix and @albertcthomas @glemaitre for testing !

PR #226, opened 2020-03-10
Sanitize user input in workflows

rth:
Adds very light input sanitizing of user submissions to make aware that poking around the RAMP server is not recommended.

For sensitive discussions please use email.

The actual long term solution is https://github.com/paris-saclay-cds/ramp-board/pull/374 but it's still useful to detect attempts and make users aware I think. Docker is not completely fullproof either https://resources.whitesourcesoftware.com/blog-whitesource/top-5-docker-vulnerabilities

### DIFF
--- a/rampwf/utils/importing.py
+++ b/rampwf/utils/importing.py
@@ -16,7 +16,7 @@ def import_module_from_source(source, name, sanitize=False):
     name : str
         Name to give to the module once loaded.
     sanitize: bool, default=False
-        check for blacklisted key words in code
+        Check for blacklisted key words in code.
 
     Returns
     -------
--- a/rampwf/utils/importing.py
+++ b/rampwf/utils/importing.py
@@ -3,9 +3,10 @@ Utility to import local files from the filesystem as modules.
 
 """
 import importlib
+from .sanitize import _sanitize_input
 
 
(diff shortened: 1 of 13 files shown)


rampwf/utils/importing.py
```diff
@@ -3,9 +3,10 @@
 
 """
 import importlib
+from .sanitize import _sanitize_input
 
 
-def import_module_from_source(source, name):
+def import_module_from_source(source, name, sanitize=False):
```
rth: Can't make `True` the default, as it's also used in various notebooks to load `problem.py`

rth:
Great. If there is no objection can someone merge then?

rampwf/utils/importing.py
```diff
@@ -14,12 +15,18 @@ def import_module_from_source(source, name):
         Path to the Python source file which will be loaded as a module.
     name : str
         Name to give to the module once loaded.
+    sanitize: bool, default=False
+        check for blacklisted key words in code
```
glemaitre:
```suggestion
        Check for blacklisted key words in code.
```

glemaitre:
Actually there is not what's new in ramp-workflow

glemaitre:
Thanks @rth

PR #230, opened 2020-03-24
Remove redundnacy in soft accuracy

### LINKED ISSUE #229: Redundancy in soft accuracy score
In [`SoftAccuracy`](https://github.com/paris-saclay-cds/ramp-workflow/blob/9805e71e68a68417c1b55c3548260b0126d88a1a/rampwf/score_types/soft_accuracy.py#L17) what is the purpose of ` y_proba = np.clip(y_proba, 0, 1)` in:
```
    def __call__(self, y_true_proba, y_proba):
        # Clip negative probas
        y_proba_positive = np.clip(y_proba, 0, 1)
        # Normalize rows
        y_proba = np.clip(y_proba, 0, 1)
        y_proba_normalized = y_proba_positive / np.sum(
            y_proba_positive, axis=1, keepdims=True)
        # Smooth true probabilities with score_matrix
        y_true_smoothed = y_true_proba.dot(self.score_matrix)
        # Compute dot product between the predicted probabilities and
        # the smoothed true "probabilites" ("" because it does not sum to 1)
        scores = np.sum(y_proba_normalized * y_true_smoothed, axis=1)
        scores = np.nan_to_num(scores)
        score = np.mean(scores)
        # to pick up all zero probabilities
        score = np.nan_to_num(score)
        return score
```
It appears to define the same thing as `y_proba_positive` and it is not used later in the function?

@agramfort @kegl

lucyleeow:
closes #229

### DIFF
--- a/rampwf/score_types/soft_accuracy.py
+++ b/rampwf/score_types/soft_accuracy.py
@@ -27,8 +27,6 @@ class SoftAccuracy(BaseScoreType):
     def __call__(self, y_true_proba, y_proba):
         # Clip negative probas
         y_proba_positive = np.clip(y_proba, 0, 1)
-        # Normalize rows
-        y_proba = np.clip(y_proba, 0, 1)
         y_proba_normalized = y_proba_positive / np.sum(
             y_proba_positive, axis=1, keepdims=True)
         # Smooth true probabilities with score_matrix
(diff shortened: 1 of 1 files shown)


rampwf/score_types/soft_accuracy.py
```diff
@@ -27,8 +27,6 @@ def __init__(self, score_matrix, name='soft precision', precision=2):
     def __call__(self, y_true_proba, y_proba):
         # Clip negative probas
         y_proba_positive = np.clip(y_proba, 0, 1)
-        # Normalize rows
```
agramfort: keep this line as it's done below

lucyleeow:
It's green, merging.

PR #233, opened 2020-05-17
Improve pickle

albertcthomas:
Pickling was working but raising a non related warning because the file was not read as a binary file. An error should have been raised instead of a warning saying that the model cannot be pickled.

The pickled file is now read as a binary file and only Pickle errors are caught by the try except clause.

PS: this is also fixing two non related flake8 errors to make Travis happy

### DIFF
--- a/rampwf/utils/submission.py
+++ b/rampwf/utils/submission.py
@@ -420,13 +420,22 @@ def pickle_model(fold_output_path, trained_workflow, model_name='model.pkl'):
     trained_workflow : a rampwf.workflow
         either the input workflow or the pickled and reloaded workflow
     """
+    msg = "Warning: model can't be pickled."
+    model_file = os.path.join(fold_output_path, model_name)
     try:
-        model_file = os.path.join(fold_output_path, model_name)
         with open(model_file, 'wb') as pickle_file:
             cloudpickle.dump(trained_workflow, pickle_file)
-        with open(model_file, 'rb') as pickle_file:
-            trained_workflow = cloudpickle.load(pickle_file)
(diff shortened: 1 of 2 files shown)


agramfort:
any chance you can add a test?

albertcthomas:
I will try to add a test. I also need to double check the errors generated when unpickling as you can have other errors than a PickleError (see [pickle doc](https://docs.python.org/3/library/pickle.html#pickle.UnpicklingError))

albertcthomas:
1. I don't know if it is possible to have a test for the original fix, replacing 'r' with 'rb' as pickle.load can raise any exception and not just Unpickling errors.

2. We can still have a test checking that if we cannot pickle then the warning message is raised.

3. With the last commit, if the trained workflow cannot be dumped we don't even try to load it, we print the warning and return the trained workflow. If it can be dumped we try to load it. The only drawback with this is that it might overcomplicate the code. Let me know what you think.

Are we sure that the load part is useful? Is it possible to have an error when loading a pickled object just after dumping it? Here we try to load it in the same module that was used to dump it, within the same python session/process. In my opinion troubles can come when trying to load a pickled object in a different session (e.g. a different python process where modules have to be imported) or with a different python version. I don't know enough about pickle for this :).

agramfort:
good to go from your end?

next time do not open a PR from the upstream repo. thx

albertcthomas:
I just rebased. If tests pass, good to go for me.

Ok for the PR from the upstream repo.

agramfort:
thx @albertcthomas

PR #237, opened 2020-06-18
[MRG] Imports and cloudpickle

### LINKED ISSUE #232: Add imported module to sys.modules
Was there a good reason for not adding ```sys.modules[name] = module``` in the [`import_module_from_source` function](https://github.com/paris-saclay-cds/ramp-workflow/blob/master/rampwf/utils/importing.py#L9) as suggested in the [python doc](https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly)?

See also issue #204 and PR #210

albertcthomas:
Fixes #232

- [x] Add a comment to `import_module_from_source` explaining why we don't add the module to `sys.modules`.

- [x] Add a test checking that `cloudpickle` works on an object that can be pickled with `cloudpickle`. Note that running `ramp-test` with `--pickle` does not really test this because the pickling errors are caught.

### DIFF
--- a/rampwf/tests/test_kits.py
+++ b/rampwf/tests/test_kits.py
@@ -2,8 +2,11 @@ import os
 import glob
 import shutil
 
+import cloudpickle
+
 import pytest
 
+from rampwf.utils import import_module_from_source
 from rampwf.utils.testing import (
     assert_submission, assert_notebook, blend_submissions)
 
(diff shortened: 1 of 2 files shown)


albertcthomas:
Hm the failing tests (which are also failing on master apparently) might be my fault because of not catching all exceptions when pickling fails in #233. There is a `TypeError` and not a `PicklingError`. Strange that it worked at the time of the PR.

albertcthomas:
Now we have a failing test because `pip install keras` in the `testing-requirement.txt` installs the last version Keras 2.4 which requires tensorflow 2.2, whereas tensorflow 2.0.1 is specified in `testing-requirement.txt`.

agramfort:
update testing-requirement.txt. ?

>

albertcthomas:
Done in PR #240

albertcthomas:
CIs are green after rebasing on master. No need to change the error handling when pickle fails for now.

albertcthomas:
@glemaitre maybe you would like to have a quick look at this PR if you have some time.

PR #242, opened 2020-07-22
Setting display.width to None

gabriel-hurtado:
Is there a reason why we print score with a fixed width of 160 [here]( https://github.com/paris-saclay-cds/ramp-workflow/blob/master/rampwf/utils/pretty_print.py#L57 ) ?

Setting `display.width` to 0 would allow pandas to autodetects the size of terminal window, avoiding useless shortening in the printed scores

### DIFF
--- a/rampwf/utils/pretty_print.py
+++ b/rampwf/utils/pretty_print.py
@@ -54,7 +54,7 @@ def print_df_scores(df_scores, indent=''):
     indent : str, default=''
         indentation if needed
     """
-    with option_context("display.width", 160):
+    with option_context("display.width", None):
         df_repr = repr(df_scores)
     df_repr_out = []
     for line, color_key in zip(df_repr.splitlines(),


agramfort:
I suspect it's also used in the web frontend. Can you check?

gabriel-hurtado:
They seem to use a different technique in the fronted [https://github.com/paris-saclay-cds/ramp-board/blob/4d908711ea25a977baa690292af31655887b42ff/ramp-database/ramp_database/tools/leaderboard.py#L20](https://github.com/paris-saclay-cds/ramp-board/blob/4d908711ea25a977baa690292af31655887b42ff/ramp-database/ramp_database/tools/leaderboard.py#L20)

agramfort:
then I don't know why it's like this. git blame will tell you who wrote this

>

gabriel-hurtado:
It seems that it was introduced with [this](https://github.com/paris-saclay-cds/ramp-workflow/pull/89) pull request

agramfort:
ask @rth he may know then

>

glemaitre:
I just assume that 160 characters were considered enough :)
I was checking the documentation. Do you want `0` or set it to `None`: https://pandas.pydata.org/pandas-docs/stable/user_guide/options.html

gabriel-hurtado:
Indeed, it should be `display_width` None rather than 0 (the undocumented option of setting it to 0 works  because in the end pandas returns `display_width or terminal_width`

rth:
As far as I remember we wanted to avoid line breaks if the score table was a bit larger than the terminal screen size. In which case scrolling horizontally might still be more readable than line breaks..  So for instance if I have a 80char wide terminal window, with None, it would break lines I think for anything larger than it?

gabriel-hurtado:
Indeed, with `None`, uses line breaks if printing more than what can fit. 
However with `display.width=160` and a terminal smaller, let's say 80, the current behavior is not to create scrolling, it is to truncate the output dataframe with `...`
Not sure what is the best way to handle it.

rth:
Aww I wasn't aware of that. Might depend on the terminal. I'm not opposed to having None there as well..

kegl:
Any objection to merging this PR?

rth:
Let's merge it then.

PR #243, opened 2020-08-19
Decoupled cv

### LINKED ISSUE #184: [RFC] Decoupling the CV object of the main script
In complex workflows (for example in transfer learning or cascades) it may happen that `train_is` and `valid_is` that `problem.get_cv` yields are more complex than just a set of indices, and that the interaction between the data (returned by `problem.get_train_data`) and the CV fold object is more elaborate than just simple indexing. Furthermore, we may want to have different operations in `problem.workflow.test_submission` at training+validation time and test time.

1. At this time the [RAMP functional data flow](https://github.com/paris-saclay-cds/ramp-workflow/files/5089942/RAMP.functional.data.flow.pdf) contains an indexing step in the [main script](https://github.com/paris-saclay-cds/ramp-workflow/blob/b92ca38287c94124444f51f41d9231d0a94db3cc/rampwf/utils/submission.py#L185), between returning from `problem.workflow.test_submission` and entering into `Predictions.__init__`. The first change will be to pass the fold indices directly to `Predictions.__init__` and rewrite all existing Predictions to the default indexing. Then complex workflows can implement other interactions between the data and the CV fold in the customizable constructor.
2. We will optionally pass `fold=(train_is, valid_is)` to `problem.workflow.test_submission`. To keep backward compatibility, the main script will accept both `problem.workflow.test_submission(trained_workflow, X)` and `problem.workflow.test_submission(trained_workflow, X, fold)`.

The changes will be fully backward compatible, and will happen under the lid.

The new simplified and fully customizable [RAMP functional data flow](https://github.com/paris-saclay-cds/ramp-workflow/files/5089959/RAMP.functional.data.flow.1.pdf) reflects what will happen after the changes.

kegl:
This is to close #184.
- [x] getting rid of indexing in main script to allow flexible cross validation
- [x] passing fold object to `test_submission`
- [x] getting rid of the necessity of indexing in the scores
- [x] adding the data flow to the documentation


agramfort:
@kegl do you anticipate any issue with ramp-board when updating rampwf based on this PR?

kegl:
@agramfort no, the interface remained intact. 

The only issue is if a kit implemented its own Predictions type: in this case __init__ needs to be changed. AFAIK there is no such kit, but I think the best tool for this would be https://travis-ci.org/github/paris-saclay-cds/rampwf-kits-test-master: all kits that we want to maintain should be there.

kegl:
While editing scoring.rst, I corrected a small mistake: the public leaderboard displays the *bagged* validation score, not the mean. I checked it on the new ramp.studio site, it still seems to be the case.

agramfort:
if there is a bug on master can you fix it in a dedicated PR? thanks

>

kegl:
It was a mistake in the documentation (not in the code) which I edited anyway as part of documenting the data flow.

kegl:
@agramfort is it OK if @albertcthomas and/or @gabriel-hurtado reviews this or would you like someone also from the CDS side?

agramfort:
I just want to guarantee that it will not break the ramp.studio website if
we use what comes out
of this PR.

>

kegl:
Yes, I agree. How can we check that? Can we do a branch on ramp-board where we change travis and pull ramp-workflow from this branch instead of master?

kegl:
@agramfort would my previous suggestion assure you? If yes, could pls @maikia do it, I'm not very skillful at scripting the new test. If not, what would be your suggestion? How did you test previous updates to ramp-workflow?

kegl:
I added some scripts to the doc that can be executed line by line e.g. in a notebook, e.g for  debugging a RAMP kit. Furthermore it can also play a pedagogic role to show ML students what single and double cross validation are. cc @agramfort https://github.com/paris-saclay-cds/ramp-workflow/blob/9d9ccedf81286cb2d85c9415c2900e1639df3bfd/doc/scoring.rst

kegl:
I added the general test protocol to test ramp-board with the PR'd ramp-workflow branch to https://github.com/kegl/ramp-workflow/blob/decoupled_cv/README.rst. I followed the protocol to test this PR on ramp-board. There was an issue since ramp-board does not completely rely on the ramp-workflow scoring. I fixed it and make all tests green. Big shout-out to @glemaitre and @jorisvandenbossche for writing the ramp-board tests, without those these kinds of changes would have been very dangerous. Thanks!

I think now the PR can be merged, let me know @agramfort if we can go ahead.

PR #247, opened 2020-09-15
Fixing ramp-leaderboard

kegl:


agramfort:
@kegl please do not open PRs from upstream repo.

We need to keep only the good branches in https://github.com/paris-saclay-cds/ramp-workflow/branches

PR #252, opened 2020-10-23
FIX encoding for new nbconvert version

albertcthomas:
Appveyor is red now that nbconvert is installed with a version higher than 6. It was green when using a version < 6.

Trying to fix it by forcing utf8 encoding when reading the notebook. Note that the notebook is written to a file with an utf8 encoding (https://github.com/paris-saclay-cds/ramp-workflow/blob/master/rampwf/utils/notebook.py#L59)


agramfort (approved):

agramfort:
thx @albertcthomas

PR #261, opened 2021-02-11
Autocompletion for submission names

albertcthomas:
If this can be useful to anyone I am sharing this here.

This was done following [click documentation](https://click.palletsprojects.com/en/7.x/bashcomplete/#shell-completion).

As written in the [documentation](https://click.palletsprojects.com/en/7.x/bashcomplete/#activation) I need to activate the autocompletion for my shell for this to work.

### DIFF
--- a/doc/command_line.rst
+++ b/doc/command_line.rst
@@ -3,6 +3,13 @@
 RAMP-workflow commands
 ######################
 
+The following commands are built using the
+`click <https://click.palletsprojects.com/en/7.x/>`_ package which provides tab
+completion for the command options. You however need to activate shell
+completion by following the instructions given in the `click documentation <https://click.palletsprojects.com/en/7.x/bashcomplete/#activation>`_.
+The `ramp-test` command also comes with tab completion for the submission name
+if the submission you are looking for is located in the `./submissions/` folder.
+
 .. click:: rampwf.utils.cli.testing:main
     :prog: ramp-test
     :show-nested:
--- a/rampwf/utils/cli/testing.py
+++ b/rampwf/utils/cli/testing.py
@@ -9,11 +9,31 @@ from ..testing import assert_submission
 CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
 
 
+def get_submissions(ctx, args, incomplete):
+    """Call-back for submission autocomplete.
+
+    Find the possible submission names and use them for autocompletion.
+    This only works if the submissions can be found in a 'submissions' folder.
+    """
+    submission_dir_path = os.path.join('.', 'submissions')
+    try:
+        all_submissions = [
+            f.name for f in os.scandir(submission_dir_path)
+            if f.is_dir() and f.name != '__pycache__']
+    except FileNotFoundError:
+        return []
+    else:
+        valid_submissions = [submission for submission in all_submissions
+                             if incomplete in submission]
+        return valid_submissions
+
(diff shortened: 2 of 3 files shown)


kegl:
Amazing, thanks!

Could also be done for the data label, examining folders in ./data

@agramfort do you or someone from Saclay wants to approve it?

rth (commented):
This looks nice indeed!

Could you please add,
>  I need to activate the autocompletion for my shell for this to work.

somewhere in the ramp-workflow documentation?

albertcthomas:
> Could also be done for the data label, examining folders in ./data

Yes in the advanced branch.

albertcthomas:
> Could you please add,
> 
> > I need to activate the autocompletion for my shell for this to work.
> 
> somewhere in the ramp-workflow documentation?

Yes, this could also benefit from some tests, if not the full autocompletion thing, which can be hard to test, maybe the function itself.

albertcthomas:
I also tried something using the `args` argument of the `get_submissions` function so that we could look if `ramp-submission-dir` is specified and look into this directory. However I don't really know this path is specified in the command line and trying it lead to several errors that I did not want to deal with to not make a function hard to test too complicated. Besides is there a real use of this option besides on the server running the submissions submitted by the users and in the tests where no autocompletion is needed? It seems I would prefer telling the users to call `ramp-test` in the root dir where they have the `problem.py` and the `submissions` folder. Anyway this is maybe going out of the scope of this PR :)

agramfort:
LGTM 

+1 for merge if CIs are green

albertcthomas:
> > Could you please add,
> > > I need to activate the autocompletion for my shell for this to work.
> > 
> > 
> > somewhere in the ramp-workflow documentation?
> 
> Yes, this could also benefit from some tests, if not the full autocompletion thing, which can be hard to test, maybe the function itself.

Actually, `ramp-test` is not in the documentation, only `ramp_test_submission` which does not use `click`...

agramfort:
do you have a minute to update the doc? thx

no objection to merge here

albertcthomas:
> do you have a minute to update the doc? thx

I am not sure what the plan is here. Should we just replace `ramp_test_submission` by `ramp-test`? Add a `ramp-test` entry? What was the reason to keep both? Backward compatibility?

agramfort:
I would just document ramp-test now. ramp_test_submission should ideally be
deprecated

>

albertcthomas:
I’ll document `ramp-test` and leave the rest for another PR

albertcthomas:
`ramp-test` was in fact already in the master doc (but not in the doc of the last release).

kegl:
I took care of this already, the old cli is deleted from the advanced branch:
https://github.com/paris-saclay-cds/ramp-workflow/blob/advanced/setup.py
https://paris-saclay-cds.github.io/ramp-docs/ramp-workflow/advanced/command_line.html

We should merge it into master.

albertcthomas:
It's already in master (https://github.com/paris-saclay-cds/ramp-workflow/blob/master/doc/command_line.rst). I got confused for not looking at the correct version: the documentation website shows the documentation of the last stable release (0.3.3), not the one of master (0.4).

agramfort (approved):

agramfort:
Thx @albertcthomas

albertcthomas:
Thanks for the reviews and comments :)

PR #262, opened 2021-02-12
Fixing CI

albertcthomas:
~First trying with scikit-learn < 0.24 before investigating a bit more.~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: python
-dist: trusty
+dist: xenial
 cache: pip
 branches:
   only:


albertcthomas:
The failure can be fixed by upgrading to ubuntu xenial (16.04), see for instance [here](https://discuss.mxnet.apache.org/t/import-mxnet-throws-importerror-dlopen-cannot-load-any-more-object-with-static-tls/3318/9) for a related issue. As Xenial is now the [min-dependency assumed for scikit-learn](https://github.com/scikit-learn/scikit-learn/pull/12746#issuecomment-445817628) this might be a good enough fix here... WDYT @agramfort?

agramfort:
Thx @albertcthomas

PR #266, opened 2021-04-15
MAINT/Fix security vulnerability warning by updating to tensorflow 2.0.4

rth:
Fixes Github security warnings on the home page (and in email reminders). Tensorflow can probably be updated to a later version as well but I'm not fully sure about the implications and for now I just want to remove the warnings.

### DIFF
--- a/testing-requirements.txt
+++ b/testing-requirements.txt
@@ -9,6 +9,5 @@ netCDF4
 xarray
 scikit-image
 joblib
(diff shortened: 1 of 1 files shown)


rth:
Hmm, it looks like this update makes drug_spectra training fail on Python 3.6 only.

agramfort:
bump drug_spectra to 3.7?

>

rth:
No actually, drug_spectra test is just flaky including on master. Re-starting the CI fixed it. We need to address that but it's not related to this PR.

Issue #264, opened 2021-04-15
Release 0.4.0


rth:
Last release was in March 2020, and I think it would be good to make a new release before trying to merge the advanced branch.

There wasn't really much breaking changes since 0.3.3 but one year later 0.4.0 is still probably better than 0.3.4  

TODO: 

- [x] update changelog
- [x] likely include https://github.com/paris-saclay-cds/ramp-workflow/pull/245 

The other more major changes in https://github.com/paris-saclay-cds/ramp-workflow/pull/246 with potential backward compatibility implications could be included in a follow up 0.5.0 release

@glemaitre One everything is ready would you be able to upload it to PyPi? Otherwise I can do it, but I would need permissions (PyPi username: ryurchak).

glemaitre:
Once this is ready, I can find time to make the upload to PyPI

rth:
OK, I created the 0.4.0 tag and if I didn't mess up anything it should be good to go,
```
twine check dist/ramp_workflow-0.4.0-py3-none-any.whl 
Checking dist/ramp_workflow-0.4.0-py3-none-any.whl: PASSED, with warnings
  warning: `long_description_content_type` missing. defaulting to `text/x-rst`.
```
@glemaitre Could you also please change the symlink for the docs https://paris-saclay-cds.github.io/ramp-docs/ramp-workflow/dev/contribute.html#release-process ? Thanks!

glemaitre:
yep

glemaitre:
Done for both:

https://pypi.org/project/ramp-workflow/
https://paris-saclay-cds.github.io/ramp-docs/ramp-workflow/stable/

I also bump master to 0.5.0.dev0 and add the info in the "what's new"

glemaitre:
Feel free to close the issue if we are good.

rth:
Thanks @glemaitre!

PR #257, opened 2020-11-25
Advanced hyperopt

gabriel-hurtado:
Minor changes to hyperopt

PR #268, opened 2021-04-15
Sync minor improvements for 0.4.0 from the advanced branch

rth:
This aims to add to master 2 PRs from advanced branch that I think should be fairly safe to include in 0.4.0  (#264 )
 -  adding --data-label (#245) 
 - hyperopt fix (#257)

For the other more major PRs in advanced, I think it would be better to include them for 0.5.0 .

Will merge this **without** squashing to preserve history and attributions.


rampwf/utils/testing.py
```diff
@@ -37,18 +37,31 @@ def assert_title(ramp_kit_dir='.'):
...
 
 
-def assert_data(ramp_kit_dir='.', ramp_data_dir='.'):
+def assert_data(ramp_kit_dir='.', ramp_data_dir='.', data_label=None):
     problem = assert_read_problem(ramp_kit_dir)
-    print_title('Reading train and test files from {}/data ...'.format(
-        ramp_data_dir))
-    X_train, y_train = problem.get_train_data(path=ramp_data_dir)
-    X_test, y_test = problem.get_test_data(path=ramp_data_dir)
+    if data_label is None:
```
agramfort, line 42: I would be more natural to branch inside get_train_data and get_test_data

I am sure you don't do think to stay backward compatible but maybe can you add a comment to explain this?

rampwf/utils/testing.py
```diff
@@ -172,6 +197,9 @@ def blend_submissions(submissions, ramp_kit_dir='.', ramp_data_dir='.',
         The directory of the ramp-kit to be blended.
     ramp_data_dir : str, default='.'
         The directory of the data.
+    data_label : str, default=None
+        The subdirectory of data in /data and training outputs in
```
agramfort, line 201:
```suggestion
        The subdirectory of data in {ramp_data_dir}/data and training outputs in
```

rampwf/utils/testing.py
```diff
@@ -172,6 +197,9 @@ def blend_submissions(submissions, ramp_kit_dir='.', ramp_data_dir='.',
...
+        /submissions/<submission>/training_output
```
agramfort, line 202:
```suggestion
        {ramp_kit_dir}/submissions/<submission>/training_output
```

agramfort (commented):
also we should have a changelog / what's new page to document changes

thx @rth

rth:
> also we should have a changelog / what's new page to document changes

Indeed updated the changelog in https://github.com/paris-saclay-cds/ramp-workflow/pull/265, I was thinking to merge that one first to avoid to deal with merge conflicts here.

agramfort:
ok fair enough... merge when happy

>

rampwf/utils/testing.py
rth: Indeed, I simplified that this code a bit.

albertcthomas:
Thanks @rth and @agramfort

PR #240, opened 2020-06-20
[MRG] Fix CI with keras version

albertcthomas:
Fixes #239 

Specifies keras version in `testing-requirements.txt` for compatibility with current tensorflow version.

PR #271, opened 2021-04-23
Add Decoupled CV to master (#240)

rth:
This is a backport of https://github.com/paris-saclay-cds/ramp-workflow/pull/243 to master.

I'll mention https://github.com/paris-saclay-cds/ramp-board/pull/515 in the test protocol documentation which should allow running ramp-board CI on arbitrary ramp-workflow PRs using [manual triggers](https://github.com/paris-saclay-cds/ramp-board/pull/515)


albertcthomas:
I think the first 2 commits (the one different to Decoupled CV) are already on master and not needed in this PR. I don't think they appear in the diff so maybe it's not a big deal.

I also think that we could use this opportunity to remove the co-authoring of the decoupled CV? @kegl used 2 usernames.

albertcthomas:
> I don't think they appear in the diff so maybe it's not a big deal.

yes they don't appear in the changes.

rth:
> I also think that we could use this opportunity to remove the co-authoring of the decoupled CV

Thanks for checking @albertcthomas ! I removed them.

rth (commented):
OK, I have changed slightly the procedure for testing with ramp-board. Now one can go to [ramp-board github actions](https://github.com/paris-saclay-cds/ramp-board/actions/workflows/main.yml) and copy paste a pip installable URL of ramp-workflow to run CI with,
image
It's still a bit manual and it requires write permissions to ramp-board to trigger CI, but at least it doesn't require to push commits to ramp-board on a separate branch and keep that branch in sync with master as the previous procedure did.

Code wise this LGTM. I can also confirm this passes [ramp-board CI on master](https://github.com/paris-saclay-cds/ramp-board/actions/runs/796964598)

PR #272, opened 2021-04-30
CLI Remove ramp_test_submission, ramp_test_notebook, ramp_convert_notebook, ramp_leaderboard

rth:
This is a backport of part of https://github.com/paris-saclay-cds/ramp-workflow/pull/247 (currently in advanced) to master.

Which removes following rampwf.utils.command_line endpoints, 
 - `ramp_test_submission` (in favor of `ramp-test`)
 - `ramp_test_notebook`
 - `ramp_convert_notebook`
 - `ramp_leaderboard`

except for `ramp_test_submission` I imagine are somewhat redundant with rampwf.utils.cli endpoints or not useful?  As far as I can tell they are never used in ramp kits or ramp-board?

One concern is that there [are deployed ramp events that suggest to use `ramp_test_submission`](https://github.com/search?q=org%3Aramp-kits+ramp_test_submission&type=code) instead of `ramp-test` so we should probably not remove it but rather deprecate (or expose it as an alias to `ramp-test` if the API is close enough)

For the other 3 endpoints, I'm not sure if we need a deprecation cycle or removing is fine. I imagine users shouldn't have been using those?

@agramfort @glemaitre Please let me know if you have any concern about this. I haven't followed the historical reasons for having both `ramp-test` and `ramp_test_submission`.

cc @albertcthomas 

TODO:
 - [x] add changelog entry once there is an agreement on what should be done.


agramfort (approved):
i've never used ramp_test_notebook, ramp_convert_notebook, ramp_leaderboard and we have been advertising for some time now the use of ramp-test so fine with me. I am not pushing for a deprecation here.

rth:
Great, thanks for the confirmation!

kegl:
Don't we use ramp_convert_notebook when we set up the event (to convert the notebook into html and display it on the frontend)? Just a neuron firing from distant memory.

rth:
As far as I can tell in ramp-board we are now [directly using `jupyter nbcovert`](https://github.com/paris-saclay-cds/ramp-board/blob/66304d45a3e62de53bd17c0ae16982ba0ac7fa5d/ramp-utils/ramp_utils/deploy.py#L50) CLI rather than calling this script.

PR #277, opened 2021-06-02
Utils

martin1tab:


albertcthomas:
Thanks for the PR @martin1tab. Could you please add a description in the top comment and reference the associated issue?

rampwf/utils/cli/testing.py
```diff
@@ -92,7 +93,9 @@ def main(submission, ramp_kit_dir, ramp_data_dir, data_label,
         ]
     else:
         submission = [submission]
-
+    if not os.path.exists("utils"):
```
albertcthomas, line 96: I think we should append to `sys.path` only if `utils` exists, otherwise there is no need to create an empty directory.

rampwf/utils/cli/testing.py
```diff
@@ -92,7 +93,9 @@ def main(submission, ramp_kit_dir, ramp_data_dir, data_label,
...
+        os.makedirs("utils")
+    sys.path.append(os.path.join(ramp_kit_dir, "utils"))
```
albertcthomas, line 98: Wouldn't it be better to insert the path at the beginning of `sys.path`? This would mimick the fact that python first searches local modules before searching for libraries, although I am not completely sure.

albertcthomas (commented):
With this implementation I think that if the structure is the following

albertcthomas:
The failure of circle appears to be due to a change in Sphinx version.

rampwf/utils/cli/testing.py
martin1tab: OK!

rampwf/utils/cli/testing.py
martin1tab: I am not sure either, I think that is recommended to do .append()

martin1tab:
> Sphinx

I will create a new PR to solve this issue!

PR #279, opened 2021-06-03
ENH add the external_imports directory to sys.path

### LINKED ISSUE #181: [RFC] importing files in submissions
A lot of times submitters would like to reuse some local libraries they have, importing them in the models (e.g., regressor.py). This is not possible at this point. I am not sure whether we want to have this feature because it would make the code non-portable, and of course non-submittable to ramp board. But it may be an optional feature during the development of the pipeline.

martin1tab:
Closes #181

With this PR merged, users would be able to import packages copied to external_imports directory.
This PR adds the external_imports directory to sys.path.

Let's imagine this kit structure:

problem.py
data/
external_imports/
......utils/
............test_imports.py

Then to import test_imports.py, it is sufficient to do (from inside problem.py or estimator/classifier.py):
from utils import test_imports

### DIFF
--- a/doc/problem.rst
+++ b/doc/problem.rst
@@ -310,5 +310,17 @@ file. It is worth taking a look at the `whole file
    The ``_read_data()`` is not strictly required and is acting as a helper
    function in the code above.
 
+8. If the ``problem.py`` file becomes too long and you would like to refactor
+   it, you can add an ``external_imports`` folder in the ``ramp_kit_dir`` and
+   have modules there that you can import from. The way this works is that
+   running ``ramp-test`` adds the ``external_imports`` folder to ``sys.path``
+   if such a folder exists. For instance if you have a ``utils.py`` module
+   in the ``external_imports`` folder::
+
+        external_imports/
+            utils.py
+
+   Then you can do ``import utils`` in ``problem.py``.
+
 .. _RAMP workflow: https://github.com/paris-saclay-cds/ramp-workflow
 .. _Pollenating insects: <https://github.com/ramp-kits/pollenating_insects>`_
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -15,7 +15,7 @@ Added
 - Added decoupled Cross Validation (`#271 <https://github.com/paris-saclay-cds/ramp-workflow/pull/271>`_)
 - Adds an option to ramp-blend to select the ``--score_metrics_index`` to blend on
   (`#273 <https://github.com/paris-saclay-cds/ramp-workflow/pull/273>`_)
-- Adds the possibility to import modules located inside the external_imports directory
+- Adds the possibility to import modules located inside the ``external_imports`` directory
   (`#181 <https://github.com/paris-saclay-cds/ramp-workflow/pull/279>`_)
 
 Changed
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -16,7 +16,7 @@ Added
 - Adds an option to ramp-blend to select the ``--score_metrics_index`` to blend on
   (`#273 <https://github.com/paris-saclay-cds/ramp-workflow/pull/273>`_)
 - Adds the possibility to import modules located inside the ``external_imports`` directory
-  (`#181 <https://github.com/paris-saclay-cds/ramp-workflow/pull/279>`_)
+  (`#279 <https://github.com/paris-saclay-cds/ramp-workflow/pull/279>`_)
 
 Changed
 -------
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -15,6 +15,8 @@ Added
 - Added decoupled Cross Validation (`#271 <https://github.com/paris-saclay-cds/ramp-workflow/pull/271>`_)
 - Adds an option to ramp-blend to select the ``--score_metrics_index`` to blend on
   (`#273 <https://github.com/paris-saclay-cds/ramp-workflow/pull/273>`_)
+- Adds the possibility to import modules located inside the external_imports directory
+  (`#181 <https://github.com/paris-saclay-cds/ramp-workflow/pull/279>`_)
 
 Changed
 -------
--- a/rampwf/utils/testing.py
+++ b/rampwf/utils/testing.py
@@ -5,6 +5,7 @@ from __future__ import print_function
 
 import os
 import shutil
+import sys
 
 import numpy as np
 import pandas as pd
@@ -26,6 +27,10 @@ def assert_notebook(ramp_kit_dir='.'):
 
 
 def assert_read_problem(ramp_kit_dir='.'):
+    # allowing external imports from the external_imports folder if it exists
+    ext_imp_dir = os.path.join(ramp_kit_dir, "external_imports")
+    if os.path.exists(ext_imp_dir) and ext_imp_dir not in sys.path:
+        sys.path.append(ext_imp_dir)
     # giving a random name to the module so it passes looped tests
     return import_module_from_source(
         os.path.join(ramp_kit_dir, 'problem.py'), 'problem'
--- a/rampwf/tests/test_kits.py
+++ b/rampwf/tests/test_kits.py
@@ -35,10 +35,39 @@ def _generate_grid_path_kits():
     return grid
 
 
+def test_external_imports(tmpdir):
+    path_kit = tmpdir.join("titanic_external_imports")
+    shutil.copytree(os.path.join(PATH, "kits", "titanic"), path_kit)
+    problem_path = os.path.join(path_kit, "problem.py")
+    submissions_dir = os.path.join(path_kit, 'submissions')
+    submission_path = os.path.join(submissions_dir, 'starting_kit')
+    estimator_path = os.path.join(submission_path, "estimator.py")
+    for path in [problem_path, estimator_path]:
+        with open(path, 'r+') as f:
+            content = f.read()
+            f.seek(0, 0)
+            line1 = "from utils import test_imports as test"
+            line2 = "assert test.x == 2"
+            f.write(line1 + '\n' + line2 + '\n' + content)
+    utils_path = path_kit.mkdir("external_imports").mkdir("utils")
+    with open(os.path.join(utils_path, "test_imports.py"), 'w+') as f:
+        content = f.read()
+        f.seek(0, 0)
+        line = "x = 2"
+        f.write(line + '\n' + content)
+    assert_submission(
+        ramp_kit_dir=path_kit,
+        ramp_data_dir=path_kit,
+        ramp_submission_dir=submissions_dir,
+        submission=submission_path,
+        is_pickle=True,
+        save_output=False,
+        retrain=True)
+
+
 @pytest.mark.parametrize(
     "path_kit",
-    _generate_grid_path_kits()
-)
+    _generate_grid_path_kits())
 def test_notebook_testing(path_kit):
     # check if there is a notebook to be tested
     if len(glob.glob(os.path.join(path_kit, '*.ipynb'))):
--- a/rampwf/tests/test_kits.py
+++ b/rampwf/tests/test_kits.py
@@ -1,6 +1,7 @@
 import os
 import glob
 import shutil
+from textwrap import dedent
 
 import cloudpickle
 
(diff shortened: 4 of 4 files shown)


albertcthomas:
Please stay in the original PR #277 and change the code in the original PR.

martin1tab:
> Please stay in the original PR #277 and change the code in the original PR.

You are right, sorry I have lost my fork, so could not open my original PR

rampwf/utils/cli/testing.py
```diff
@@ -92,7 +93,8 @@ def main(submission, ramp_kit_dir, ramp_data_dir, data_label,
         ]
     else:
         submission = [submission]
-
+    if os.path.exists("utils"):
+        sys.path.append(os.path.join(ramp_kit_dir))
```
albertcthomas, line 97: you added this because of my [comment in the previous PR](https://github.com/paris-saclay-cds/ramp-workflow/pull/277#pullrequestreview-675022084) saying that it would seem more natural to be able to do `from utils import foo` than `import foo`. I wanted to add to my comment that this would make us add the `ramp_kit_dir` in `sys.path` and that this was maybe not what we wanted either. So maybe `import foo`, although less natural would be better.

albertcthomas:
After discussing with @kegl and @avirmaux in real life, if we put the `ramp_kit_dir` in `sys.path` then we don't control what users will be able to import, they'll have access to everything in the `ramp_kit_dir` which could be bad (correct me if I'm wrong). Thus we keep the idea of the folder where you are able to do imports, this folder could be named `external_imports` and would contain `utils/` or any other folder and by adding `external_imports` to `sys.path` you would be able to do `from utils import foo`. It's a bit more obscure for a python user (maybe) but the name of the folder and documentation would help.
WDYT @rth? @martin1tab?

rth:
>  Thus we keep the idea of the folder where you are able to do imports, this folder could be named external_imports and would contain utils/ or any other folder and by adding external_imports to sys.path you would be able to do from utils import foo

This could work indeed as long as it's documented.

Though in the end wouldn't be simpler to document that people can set [`PYTHONPATH`](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPATH) to add arbitrary folders to the Python import path?
```
PYTHONPATH="/some_extra_packages/"  ramp-test
```
(or when starting the notebook)

That would be a fairly standard solution with fewer constraints about hard-coded folder paths. Unless it wouldn't work for some reason?

kegl:
I'd prefer adding a standard `external_imports` so when we deploy the RAMP kit at the backend, we don't need to do any extra steps, ramp-test should work out of the box.

kegl:
I'd prefer to keep `external_imports` so we don't need additional steps (adding path) when we deploy a new challenge at the backend.

rampwf/tests/kits/titanic_test_external_imports/submissions/starting_kit/estimator.py
```diff
@@ -0,0 +1,26 @@
+from sklearn.compose import make_column_transformer
+from sklearn.preprocessing import OneHotEncoder
+from sklearn.impute import SimpleImputer
+from sklearn.pipeline import Pipeline
+from sklearn.linear_model import LogisticRegression
+from utils import test_imports as test
```
albertcthomas, line 6: I would not use the `as test` and just use `test_imports` to keep the explicit name

albertcthomas: also please add a blank line between this import and the previous ones

rampwf/tests/kits/titanic_test_external_imports/problem.py
```diff
@@ -0,0 +1,57 @@
+import os
+
+import pandas as pd
+from sklearn.model_selection import StratifiedShuffleSplit
+import sys
+import rampwf as rw
+
+from utils import test_imports as test
```
albertcthomas, line 8: see my comment below w.r.t to the naming

rampwf/tests/kits/titanic_test_external_imports/problem.py
```diff
@@ -0,0 +1,57 @@
...
+
+assert test.x == 2
```
albertcthomas, line 10: see my comment below about adding a comment

doc/problem.rst
```diff
@@ -310,5 +310,8 @@ file. It is worth taking a look at the `whole file
    The ``_read_data()`` is not strictly required and is acting as a helper
    function in the code above.
 
+8.  if problem.py becomes too long, one can have external_imports folder,
```
albertcthomas, line 313: Let's give more details, such as

8. If the ``problem.py`` file becomes too long and you would like to refactor it, you can add an ``external_imports`` folder in the ``ramp_kit_dir`` and have modules there that you can import from. The way this works is that running ``ramp-test`` adds the ``external_imports`` folder to ``sys.path`` if such a folder exists. For instance if you have
```
    external_imports/
        utils.py
```
Then you can do ``import utils`` in ``problem.py``.

rampwf/utils/testing.py
```diff
@@ -27,6 +28,9 @@ def assert_notebook(ramp_kit_dir='.'):
 
 def assert_read_problem(ramp_kit_dir='.'):
     # giving a random name to the module so it passes looped tests
```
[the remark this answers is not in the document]
albertcthomas: maybe also add a comment here to explain why we are doing this, for instance
```
# allowing external imports from the external_imports folder if it exists
```

albertcthomas (changes requested):
Thanks @martin1tab ! I am still hesitating between doing `from external_folders.utils import x` for `from utils import x`. The former would make it more explicit that we are using an external import but puts the whole `ramp_kit_dir` in `sys.path`, still not sure if this is an issue.

rampwf/tests/kits/titanic_test_external_imports/external_imports/utils/test_imports.py
```diff
@@ -0,0 +1 @@
+x = 2
```
albertcthomas, line 1: also maybe add a small comment here to say that this is just to test external imports

rth:
Could we do without adding `rampwf/tests/kits/titanic_test_external_imports` folder with all the data files? For testing it you could for instance create a [tempdir](https://docs.pytest.org/en/6.2.x/tmpdir.html), copy one of the existing kits from `rampwf/tests/kits` there and add the file you need. Wouldn't that work?

kegl:
@albertcthomas If we also add the `--import` option to `ramp-test`, one can use it to add `.` in a local repo not intended for a data challenge, and have `utils` in the root kit folder, do `from utils import ...`. Then in case we want to turn it into a challenge, we can simply copy `utils` into `external_imports` sor`ramp-test` with no option works.

doc/problem.rst
```diff
@@ -310,5 +310,14 @@ file. It is worth taking a look at the `whole file
    The ``_read_data()`` is not strictly required and is acting as a helper
    function in the code above.
 
+8. If the problem.py file becomes too long and you would like to refactor it,
```
albertcthomas, line 313: @martin1tab this needs to be formatted correctly: use double backticks for ``problem.py`` (` ``problem.py`` `). Same for the name of the folders ``external_imports`` and ``ramp_kit_dir``, ``ramp-test``, ``sys.path``. We also need a good format for the example, making sure we have the correct indentation.

rampwf/tests/kits/titanic_test_external_imports/problem.py
```diff
@@ -0,0 +1,58 @@
+import os
+
+import pandas as pd
+from sklearn.model_selection import StratifiedShuffleSplit
+import sys
+import rampwf as rw
+
+from utils import test_imports
+
+#x is here to test external imports
```
albertcthomas, line 10:
```suggestion
# testing external imports
```

rampwf/tests/kits/titanic_test_external_imports/submissions/starting_kit/estimator.py
```diff
@@ -0,0 +1,28 @@
+from sklearn.compose import make_column_transformer
+from sklearn.preprocessing import OneHotEncoder
+from sklearn.impute import SimpleImputer
+from sklearn.pipeline import Pipeline
+from sklearn.linear_model import LogisticRegression
+
+from utils import test_imports
+
+# x is here to test external imports
```
albertcthomas, line 9:
```suggestion
# testing external imports
```

albertcthomas (changes requested):

martin1tab:
> Could we do without adding `rampwf/tests/kits/titanic_test_external_imports` folder with all the data files? For testing it you could for instance create a [tempdir](https://docs.pytest.org/en/6.2.x/tmpdir.html), copy one of the existing kits from `rampwf/tests/kits` there and add the file you need. Wouldn't that work?

OK !!

albertcthomas:
> @albertcthomas If we also add the `--import` option to `ramp-test`, one can use it to add `.` in a local repo not intended for a data challenge, and have `utils` in the root kit folder, do `from utils import ...`.

yes that would work but for now instead of adding new options and changing `sys.path` users can use `python -m`.

> I am still hesitating between doing `from external_imports.utils import x` and `from utils import x`.

I was just thinking that the former was more explicit and that it would be easier for a new user to know where the import is coming from. However if someone was using `python -m` (outside a challenge use) and had modules in the root kit dir, copying them to `external_imports` would also require to change the imports to add `external_imports`.

Maybe users should not use `python -m` and put all their external imports in `external_imports` :)

doc/problem.rst
```diff
@@ -310,5 +310,14 @@ file. It is worth taking a look at the `whole file
    The ``_read_data()`` is not strictly required and is acting as a helper
    function in the code above.
 
+8. If the  ``problem.py`` file becomes too long and you would like to refactor it, 
+   you can add an external_imports folder in the ``ramp_kit_dir`` and have modules 
+   there that you can import from. The way this works is that running ``ramp-test`` 
+   adds the ``external_imports`` folder to ``sys.path`` if such a folder exists. 
+   For instance if you have
+   external_imports/
```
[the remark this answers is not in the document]
albertcthomas: you can check the rendering here https://396-91170031-gh.circle-artifacts.com/0/doc/_build/html/stable/problem.html

doc/problem.rst
```diff
@@ -310,5 +310,14 @@ file. It is worth taking a look at the `whole file
    The ``_read_data()`` is not strictly required and is acting as a helper
    function in the code above.
 
+8. If the  ``problem.py`` file becomes too long and you would like to refactor it, 
+   you can add an external_imports folder in the ``ramp_kit_dir`` and have modules
```
albertcthomas, line 314:
```suggestion
   you can add an ``external_imports`` folder in the ``ramp_kit_dir`` and have modules 
```

doc/problem.rst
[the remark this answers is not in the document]
albertcthomas: I fixed the rendering

albertcthomas (approved):

albertcthomas:
I did a bit of cleaning. If tests pass this is good for me, thanks @martin1tab

albertcthomas:
@kegl @rth please merge if this is good for you as well

rth:
Could you please add a what's new entry https://github.com/paris-saclay-cds/ramp-workflow/blob/master/doc/whats_new.rst#050-unreleased?

martin1tab:
> Could you please add a what's new entry https://github.com/paris-saclay-cds/ramp-workflow/blob/master/doc/whats_new.rst#050-unreleased?

Yes

doc/whats_new.rst
```diff
@@ -15,6 +15,8 @@ Added
 - Added decoupled Cross Validation (`#271 <https://github.com/paris-saclay-cds/ramp-workflow/pull/271>`_)
 - Adds an option to ramp-blend to select the ``--score_metrics_index`` to blend on
   (`#273 <https://github.com/paris-saclay-cds/ramp-workflow/pull/273>`_)
+- Adds the possibility to import modules located inside the ``external_imports`` directory
+  (`#181 <https://github.com/paris-saclay-cds/ramp-workflow/pull/279>`_)
```
albertcthomas, line 19: I think it should be
```suggestion
  (`#279 <https://github.com/paris-saclay-cds/ramp-workflow/pull/279>`_)
```

martin1tab: yes, sure

rth (approved):
Thanks @martin1tab @albertcthomas !

PR #282, opened 2021-06-23
Fix blending decoupled cv

kegl:
Porting the blending code to decoupled cv.


rth (approved):
Thanks! So this file has a 100% code coverage, but it looks like this wasn't caught by the test suite.

It might be good to add a corresponding test, maybe not in this PR?

kegl:
Yes, now that we can have nontrivial CVs, it would be good to have one among the starting kits (where folds are not pairs of indices but more complex objects).

kegl:
> It might be good to add a corresponding test, maybe not in this PR?

We are planning to have a transfer learning workflow/score/Predictions, Predictions will have a set of source and target sets, at that point we will have a unit test for the nontrivial CV.

PR #287, opened 2021-11-02
Partial training

albertcthomas:
On behalf of @kegl 

This PR adds the possibility of partial training to the workflow by loading a previous pickled model
- [x] add command-line option `--partial-train`
- [x] add optional prev_model to train_workflow in classifier, regressor, feature_extractor_classifier, and feature_extractor_regressor
- [x] titanic_old test with partial training
- [x] air_passenger_old test with partial training
- [x] decouple pickling and unpickling
- [x] test pickle_model and unpickle_model
- [x] name change from (un)pickle_model to (un)pickle_trained_workflow to be more precise

PR #288, opened 2021-11-05
CI Migrate to Github Actions

rth:
Since TavisCI.org no longer works https://github.com/paris-saclay-cds/ramp-workflow/pull/287#issuecomment-959005764


kegl:
Thx a lot @rth !

.github/workflows/main.yml
```diff
@@ -0,0 +1,68 @@
...
+  push:
+    branches: [master, gh-actions-with-upstream-branch]
+  pull_request:
+    branches: [master]
+
+jobs:
+  test:
+    strategy:
+      matrix:
+        python_version: [3.6, 3.7]
```
rth, line 13: I can't use much newer versions, because the tensorflow version we install doesn't have wheels for Py3.9 for instance. This can be updated in a separate PR.

rth (commented):
The CI should now be green, however someone should disable appveyor as it's still triggering even without a config file and it looks like I don't have permissions to disable it.

agramfort (approved):

kegl:
Ok for me, thx a lot @rth, let's merge.

PR #296, opened 2021-12-16
Fix new typo found by codespell

DimitriPapadopoulos:


agramfort (approved):

agramfort:
I don't know why CIs are red though

kegl:
Seems in sklearn PCA from the logs. Also two other tests were cancelled.

PR #299, opened 2022-01-13
Fix typos (repeated words)

DimitriPapadopoulos:
Not sure about `joined back the the` → `joined back to the` here:
> This function returns a vector of labels (cluster assignments) which is then joined back the the event id column and returned (to be passed into `prediction_types.Clustering` and evaluated by `score_types.clustering_efficiency`).


agramfort (approved):
thx @DimitriPapadopoulos

PR #303, opened 2022-03-09
ENH allow open in user code for non-os moudles

tomMoral:
For some challenges, users need to load images with Image.open
from the PIL library. This is not possible as _sanitize_input
is too restrictive. This commits intend to change this.


agramfort:
now we just need to make a release :)

agramfort:
and to update the server :)

glemaitre:
I make the release and you update the server :)

PR #307, opened 2022-03-31
FIX replace remaining autocompletion from click -> shell_complete

frcaud:


agramfort (approved):

rampwf/utils/cli/testing.py
```diff
@@ -36,7 +36,7 @@ def get_submissions(ctx, args, incomplete):
               help='The kit to test. It should be located in the '
               '"submissions" folder of the starting kit. If "ALL", all '
               'submissions in the directory will be tested.',
-              autocompletion=get_submissions)
+              shell_complete=get_submissions)
```
albertcthomas, line 39: This is because the argument name changed in click?

agramfort:
yes but the shell_complete argument has been there for a while so it should
not break stuff...

albertcthomas (approved):
Thanks @frcaud

PR #313, opened 2022-08-01
Hyperopts

martin1tab:


kegl:
It seems windows tests are failing because of the netcdf library, @rth do you have any suggestion what to do?

rth:
CI fails because there are no binary wheels for Python 3.6 and 3.7 for [netcdf4](https://pypi.org/project/netCDF4/#files) and it misses the headers to build it from sources (which is particularly difficult on Windows).

What we could do is either,
 - pin netcdf4 to an earlier version for now
 - or better upgrade CI to test Python 3.8+ (preferable in a separate PR ) which will not have this issue. 
 
 BTW, if possible it might be better to keep ray an [optional dependency](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html?highlight=extras#optional-dependencies) as itself it has a lot of dependencies (particularly if we don't need it to run submissions on the ramp-board server)

martin1tab:
@rth @kegl any idea on how to solve this new error ? 

import file mismatch:
imported module 'rampwf.prediction_types.tests.test_mixed_predictions' has this __file__ attribute:
  /usr/share/miniconda/envs/testenv/lib/python3.7/site-packages/rampwf/prediction_types/tests/test_mixed_predictions.py
which is not the same as the test file we want to collect:
  /home/runner/work/ramp-workflow/ramp-workflow/rampwf/prediction_types/tests/test_mixed_predictions.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules

kegl:
Do you see the error when you run pytest locally on your machine?

kegl:
> CI fails because there are no binary wheels for Python 3.6 and 3.7 for [netcdf4](https://pypi.org/project/netCDF4/#files) and it misses the headers to build it from sources (which is particularly difficult on Windows).
> 
> What we could do is either,
> 
>     * pin netcdf4 to an earlier version for now
> 
>     * or better upgrade CI to test Python 3.8+ (preferable in a separate PR ) which will not have this issue.

What I propose is to let this error stay for now. @SofianChay will work on the time series tools in the next months, one of the goals is to get rid of netcfd and replace the test with a simpler use case using csvs.

kegl:
> BTW, if possible it might be better to keep ray an [optional dependency](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html?highlight=extras#optional-dependencies) as itself it has a lot of dependencies (particularly if we don't need it to run submissions on the ramp-board server)

Awesome, that is exactly what I had in mind but don't know how to do it. Yes, ramp-hyperopt is optional, not needed for the server, so we should make ray tune import also optional. @rth Can you do this once @martin1tab fixes the new clash?

kegl:
@martin1tab We noticed that there were a lot of formatting changes unrelated to the PR (' to ", to mention one; the convention on RAMP is single quote). Pls take these off the PR.

PR #321, opened 2022-09-19
Fix CI

DimitriPapadopoulos:
* Update GitHub Actions
* Python 3.6 has reached [end-of-life](https://devguide.python.org/versions/#unsupported-versions). Use Python 3.7 instead, and use latest 3.10 too.
* New cimg images supersede the legacy CircleCI images:
  * circleci/python ⤑ [cimg/python](https://circleci.com/developer/images/image/cimg/python)
  * circleci/postgres ⤑ [cimg/postgres](https://circleci.com/developer/images/image/cimg/postgres)
* Fix docs: we need to either upgrade Sphinx, or pin down Jinja2: https://github.com/pallets/jinja/issues/1630
* Update test dependencies to fix this error:
  ```
  ERROR: Could not find a version that satisfies the requirement tensorflow==2.0.4 (from versions: 2.8.0rc0, 2.8.0rc1, 2.8.0, 2.8.1, 2.8.2, 2.8.3, 2.9.0rc0, 2.9.0rc1, 2.9.0rc2, 2.9.0, 2.9.1, 2.9.2, 2.10.0rc0, 2.10.0rc1, 2.10.0rc2, 2.10.0rc3, 2.10.0)
  ERROR: No matching distribution found for tensorflow==2.0.4
  ```
* Remove support code for Python 2 or Python 3.3. It contained bugs.

### DIFF
--- a/rampwf/externals/tabulate.py
+++ b/rampwf/externals/tabulate.py
@@ -2,18 +2,21 @@
 
 """Pretty-print tabular data."""
 
+from __future__ import print_function
+from __future__ import unicode_literals
 from collections import namedtuple
-from platform import python_version_tuple
+import sys
 import re
 import math
+import textwrap
 
 
-if python_version_tuple() >= ("3", "3", "0"):
+if sys.version_info >= (3, 3):
     from collections.abc import Iterable
 else:
     from collections import Iterable
 
-if python_version_tuple()[0] < "3":
+if sys.version_info[0] < 3:
     from itertools import izip_longest
     from functools import partial
 
@@ -27,7 +30,6 @@ if python_version_tuple()[0] < "3":
 
     def _is_file(f):
         return hasattr(f, "read")
-
 
 else:
     from itertools import zip_longest as izip_longest
@@ -53,9 +55,14 @@ try:
 except ImportError:
     wcwidth = None
 
+try:
+    from html import escape as htmlescape
+except ImportError:
+    from cgi import escape as htmlescape
+
 
 __all__ = ["tabulate", "tabulate_formats", "simple_separated_format"]
-__version__ = "0.8.6"
+__version__ = "0.8.10"
 
 
 # minimum extra space in headers
@@ -66,6 +73,9 @@ PRESERVE_WHITESPACE = False
 
 _DEFAULT_FLOATFMT = "g"
 _DEFAULT_MISSINGVAL = ""
+# default align will be overwritten by "left", "center" or "decimal"
+# depending on the formatter
+_DEFAULT_ALIGN = "default"
 
 
 # if True, enable wide-character (CJK) support
@@ -84,9 +94,9 @@ DataRow = namedtuple("DataRow", ["begin", "sep", "end"])
 #         headerrow
 #     --- linebelowheader ---
 #         datarow
-#     --- linebewteenrows ---
+#     --- linebetweenrows ---
 #     ... (more datarows) ...
-#     --- linebewteenrows ---
+#     --- linebetweenrows ---
 #         last datarow
 #     --- linebelow ---------
 #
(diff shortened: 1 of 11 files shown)


DimitriPapadopoulos:
Almost there, still an issue with docs, despite pinning down Jinja2:
```
ImportError: cannot import name 'environmentfilter' from 'jinja2'
```
Any clue? It looks like I am not updating Sphinx or pinning down Jinja2 the right way.

albertcthomas:
It seems that when doing the [sphinx<4.0 install](https://app.circleci.com/pipelines/github/paris-saclay-cds/ramp-workflow/532/workflows/0502dddb-f6dc-431c-b81f-b2f72e5b19c3/jobs/537?invite=true#step-103-381) it installs jinja2 3.1.2
```
Collecting Jinja2>=2.3
  Downloading Jinja2-3.1.2-py3-none-any.whl (133 kB)
```

albertcthomas:
> It seems that when doing the [sphinx<4.0 install](https://app.circleci.com/pipelines/github/paris-saclay-cds/ramp-workflow/532/workflows/0502dddb-f6dc-431c-b81f-b2f72e5b19c3/jobs/537?invite=true#step-103-381) it installs jinja2 3.1.2

in fact I don't know why it install sphinx < 4.0, this must be your question then :)

rth:
> Any clue? 

Probably CI broke at the time and this was a fix. It we can make it work with the latest sphinx version that would certainly be better.

DimitriPapadopoulos:
Indeed the question is why it installs sphinx < 4.0. A quick grep comes up with:

[**setup.py**](https://github.com/DimitriPapadopoulos/ramp-workflow/blob/c4faecfec95a039192f2b873138967b10f760989/setup.py#L44)
Line 44 in c4faecf
```python
    'docs': ['sphinx>4.1.0', 'sphinx_rtd_theme', 'numpydoc', 'sphinx-click', 'jinja2<3.1.0']
```

[**doc/conf.py**](https://github.com/DimitriPapadopoulos/ramp-workflow/blob/c4faecfec95a039192f2b873138967b10f760989/doc/conf.py#L28)
Line 28 in c4faecf
```python
needs_sphinx = '4.1.0'
```

It looks like these have no effect at all, something else enforces `sphinx<4.0`.

albertcthomas:
[Here](https://github.com/paris-saclay-cds/ramp-workflow/blob/a0d27ab700abbbf91dc1134bc3683eb85b1efdc9/ci_tools/circle/build_doc.sh#L16) it asks for sphinx 4.0 in `doc_build.sh`. This is what is called in CI.

albertcthomas:
> Probably CI broke at the time and this was a fix. It we can make it work with the latest sphinx version that would certainly be better.

Yes, see #280

DimitriPapadopoulos:
I will keep `sphinx<4.0` for now, and fix CI by pinning down `jinja2<3.1.0`. Then we can address the upgrade of sphinx in a different merge request.

DimitriPapadopoulos:
Note that `tensoflow` and `keras` are not pinned down any more. A single version for all version of Python doesn't seem to be working.

DimitriPapadopoulos:
Ready for review.

rampwf/externals/tabulate.py
```diff
@@ -2,18 +2,21 @@
 
 """Pretty-print tabular data."""
```
rth, line 3: I think this is from https://github.com/astanin/python-tabulate we shouldn't really migrate it ourselves, it would be better to just include a more recent version from that repo. If they didn't change the API too much.

rth (commented):
Great, thanks for doing this!

rampwf/externals/tabulate.py
DimitriPapadopoulos: Indeed, I have include the latest formal version 0.8.10 instead of modifying the code myself.

agramfort:
thx @DimitriPapadopoulos

PR #331, opened 2023-02-03
Fix CI

rth:
Fix CI following https://github.com/paris-saclay-cds/ramp-workflow/pull/321


rampwf/prediction_types/detection.py
```diff
@@ -22,11 +22,11 @@ def __init__(self, y_pred=None, y_true=None, n_samples=None,
         if y_pred is not None:
             if fold_is is not None:
                 y_pred = y_pred[fold_is]
-            self.y_pred = np.array(y_pred)
+            self.y_pred = np.array(y_pred, dtype=object)
         elif y_true is not None:
             if fold_is is not None:
                 y_true = y_true[fold_is]
-            self.y_pred = np.array(y_true)
```
rth, line 29: I'm not 100% confident this is the right fix, but the issue seems to be that for detection prediction_types the input y_true/y_pred can be some ragged array of the form,
```py
np.array([[(1, 2, 3)], None])
```
which no longer works latest numpy without dtype=object

At the same time, I don't see what else we could do with such input. The other alternative could be,
```py
self.y_pred = np.zeros(len(y_true), dtype=object)
for idx, row in enumerate(y_true):
	self.y_pred[idx] = row
```
but it feels more brittle.

rampwf/prediction_types/tests/test_detection_predictions.py
```diff
@@ -49,10 +49,10 @@ def test_combine_ignore_none():
 
     # first item: 2 out of 3 match -> confidence of 2/3
     # second item: ignore None, so 2 out of 2 match -> confidence of 2/2 = 1
-    expected = np.empty(2, dtype=object)
-    expected[:] = [[(2. / 3, 1, 1, 1)], [(1, 3, 3, 1)]]
-    assert_allclose(expected[0], y_pred_combined[0])
```
rth, line 54: Here the issue is that otherwise, y_pred_combined[0] is,
```
[array([0.6666666666666666, 1.0, 1.0, 1.0], dtype=object)]
```
and such objects can no longer be compared against an array in numpy >1.24.0. Previously with <1.24.0 it was,
```
[array([0.66666667, 1.        , 1.        , 1.        ])]
```

frcaud:
Thanks @rth !

